PR #2, opened 2017-12-12
Further bringing package up to snuff

dkent600:
Added automated test framework (Mocha+Chai)
Configured packaging for applications
Introduced  NPS for scripts (see package-scripts.js)
Located contracts to a root "contracts" folder


contracts/ExecutableInterface.json
```diff
@@ -0,0 +1,251 @@
+{
```
tibetsprague: hmm, why are we including the contracts directly in the library? that seems antithetical to the whole point of splitting into two repos. seems like arc-js should import the contracts from daostack

lib/utils.js
```diff
@@ -22,13 +22,13 @@ export function requireContract(contractName) {
   try {
     const myWeb3 = getWeb3();
 
-    const artifact = require(`../build/contracts/${contractName}.json`);
+    const artifact = require(`../contracts/${contractName}.json`);
```
tibetsprague: why not import from the daostack module?

contracts/ExecutableInterface.json
dkent600: The contracts folder *is* being fully populated by the contracts coming from daostack -- it is part of the build process.  They are not separately maintained.

The reason I added this extra folder was because I was thinking about developers might like to mess with the contracts without having to dig into node_modules, and I was aiming towards letting them have their own custom location for a set of contracts that they might use in development.

In any case, I've gone ahead and removed that local contracts folder, though I will note that in the build the user can still use an environment variable to specify an alternate location for the contracts.  I aware that that I may be jumping the gun on this, creating a solution without a problem.... (and that `requireContract` would need to know about this env variable too), and that we might want to remove that env variable too, just to keep things clean and simple, once things have settled down and we know whether we'll want it....

dkent600: @tibetsprague ok, thinking about this, I think I had it right the first time.  As I mentioned above, what I was doing was exactly what you suggested:  importing the contracts from daostack.  That is, at build time I was copying the files up from the daostack-arc package under node_modules, into a folder in the arc-js package where they can then be easily packaged for consumption by applications using arc-js.

So I think I should restore what I had.  What do you think?

tibetsprague: hmm. it still feels weird to me to copy the files into arc-js. it makes more sense to do that if we don't then commit them into the library's repo. because if they are in the repo then we have to maintain them in two different repos. i feel fine about copying them into arc-js as part of the build, but I don't really understand why that's better than just reading them from daostack. I'm also into having a config setting to point to a different contracts folder, and maybe offering an option to import and edit the contracts if desired? but again it seems like all handling of contracts and deployment of them should be happening in daostack, not arc, so if other developers want to create a new set of contracts they can clone daostack

tibetsprague: @leviadam what are your thoughts?

dkent600: I think that what you are proposing, @tibetsprague, might work if:

0) there is no explicit dependency of arc-js on daostack-arc, easy enough.
1) the user supplies the path to daostack-arc when calling `Arc.configure()` (for `requireContract`)
2) Arc-js throws an exception in `Arc.configure()` if the referenced version of daostack-arc is not compatible.

dkent600: I'm not too wild about that solution though.  It makes things more complicated for the user.

The alternative of keeping a copy of the contracts we intend to be compatible with does not seem like a matter of "maintaining" them, rather it is pulling (not transforming) the contracts we need from daostack-arc and putting them where we need them to be, with no worries for the user and less risk of errors.

tibetsprague: just to summarize more of our conversation today. i still lean towards not copying the contracts from daostack into arc, but just loading them with import or require straight from daostack. then how would testrpc deployments work? could use a package script like
“scripts”: {
    “deploy”: “daostack-arc truffle deploy”
  },

dkent600: I'm still falling on the side of encapsulating a snapshot of contract .json files in arc-js.

Benefits:
* simpler and cleaner coding for us and the users (to confirm, see below)
* minimal code footprint and bundle size for the user in production (for sure)

Costs:
The ugliness factor due to having redundant copies of the data

In relational database programming, sometimes we "flatten" the data.  This results in some redundancy and the ugliness factor, but it can be an optimal pattern.   Similar concept here.

That said, I want to more clearly understand the implications Tibet's proposal: particularly with respect to:

1) how do we handle the case where the user wants to deploy contracts to any network that is not pre-deployed by daostack-arc, most notably testrpc.  This is already implemented cleanly in my proposal, but I need to confirm how it would work with Tibet's alternative.

2) how will `requireContract` locate the contract .json files.  Again, this is already implemented cleanly in my proposal, but I need to confirm how it would work with Tibet's alternative.

I have stuff I have to deal with today, but hope to report back by end of day or tomorrow.

dkent600: # How to do custom deployment of contracts

## Use Case: Contracts coming from ds-arc

1. start testrpc
1. run the contract deploy in a ds-arc repo
1. `npm pack` ds-arc
1. in  the application, `npm install` the ds-arc tgz you just built

If one also has an arc-js repo, then steps 1 and 2 could optionally be done using two arc-js package scripts.

## Use Case: Contracts coming from ds-arc

1. start testrpc
1. run the contract deploy in a ds-arc repo
1. `npm pack` arc-js
1. in the application, `npm install` the arc-js tgz you just built

## Summary
On the surface, the two cases are virtually identical in complexity.  In either case the user must determine the correct version of the ds-arc repo to clone, and clone it correctly.

The main difference is that in the first case we are required to have ds-arc installed in the application, in the second case, we are not.  The first case has a larger app footprint, larger bundle for the app to deploy, more moving parts, increased risk.

# How does `requireContract` find custom contracts?

## In either case:
`const artifact = require(jsonPath);`

## Use case:  Contracts coming from ds-arc:

Pointing outside the arc-rs package, and a *hard-coded assumption about the ds-arc folder structure* and *assuming that ds-arc is installed sibling to arc-js in a typical NPM-style folder structure*:

`jsonPath` would be "../../daostack-arc/build/contracts/' + contractName + '.json'"

## Use Case: Contracts coming bundled in arc-js

Pointing inside the arc-js package:

`jsonPath would be '../contracts/${contractName}.json'`

In either case we might accept from the application a configuration setting that overrides the default value for jsonPath.

## Summary

Again, the two cases are virtually identical in complexity, on the surface and the main difference is that in the first case we are required to have DS-arc installed in the application, while in the second case, we are not, meaning larger app footprint, larger bundle for the app to deploy, more moving parts, increased risk.

# Bottom Line
It's tough to anticipate all the possibilities… In general, more moving parts makes me more nervous than taking a snapshot of and encapsulating the bits that we need.  In doing so we minimize potential variations and risk, and give arc-js more potential control.

In contrast, by relying at application runtime on a separate ds-arc package as the source of contracts, we simplify the arc-js build process, but I am not seeing any other clear benefit.

Either way there is not a huge difference, right now.  But again it is hard to predict all the possibilities.  As a heuristic for predicting future risk, for me, fewer moving parts beats less redundancy.

dkent600:
This has implications:  https://github.com/daostack/daostack/pull/199

leviadam:
@dkent600  I think arc must have npm/ethpm packages. 
How will arc-js read it. Will it use the package or use sub-modules, is a different question.
This is not an easy question, as it has to somehow maintain schemes management.

dkent600:
@leviadam Arc-Js already uses npm.  Alchemy uses npm together with webpack, which plays well with npm.

Will ethpm play well in this ecosystem?  That is what we have to figure out.

dkent600:
@leviadam @tibetsprague

With the recent decisions that:
1. the Arc package will not contain contract files, 
2. arc-js is responsible for all contract migration,
3. arc-js will ship with contracts pointing at mainnet, but no other networks,
4. application consumers of arc-js will be responsible for migrating contracts to testnets of their choice

... and the ramification that arc-js will now have to commit those contracts to its repo

... are there any other reasons not to approve this PR?  I'd really like to move on from it...

orenyodfat:
Is it a real need to include testrpcDB.zip in the repo ? why ?

dkent600:
@orenyodfat 

> Is it a real need to include testrpcDB.zip in the repo ? why ?

The zip file enables Travis to run automated tests against the deployed contracts.

orenyodfat:
> The zip file enables Travis to run automated tests against the deployed contracts.

The testrpc deployment can (and should) be part of the automated tests.

PR #20, opened 2018-01-01
Prettified code, tweaked some lint rules

dkent600:
No code logic changes, just reformatting changes using prettify with default config parameters.


lib/arc.js
```diff
@@ -57,18 +57,14 @@ export function configure(options) {
...
+    allContracts: contracts,
     schemes: [
-      contracts.SchemeRegistrar
-      , contracts.UpgradeScheme
-      , contracts.GlobalConstraintRegistrar
-      , contracts.ContributionReward
-    ],
-    votingMachines: [
-      contracts.AbsoluteVote
+      contracts.SchemeRegistrar,
```
leviadam: Why are the schemes under votingMachines?

dkent600: @leviadam  I think you are seeing an illusion in the diff output.  Only AbsoluteVote is under votingMachines.

.eslintignore
```diff
@@ -1,3 +1,2 @@
 node_modules/
-dist/
-package-scripts.js
+dist/
```
orenyodfat: No new line at the end of file

lib/arc.js
```diff
@@ -24,11 +24,11 @@ export function configure(options) {
   options; // for lint
   // not used at the moment:  const network = options && options.network && options.network.name ? options.network.name : 'testrpc';
   /**
-     * TODO: supply testrpc url in options?  Problem is that at this point web3 has already been set in utils
-     * so it's too late at this point to set the url.  Would need to somehow effect this initialization before
-     * any contract imports have been attempted.  Need to figure out how to export the modules above somewhere
-     * else, or separately.
-     */
+   * TODO: supply testrpc url in options?  Problem is that at this point web3 has already been set in utils
```
orenyodfat: why need testrpc url ?

lib/contributionreward.js
```diff
@@ -93,19 +102,30 @@ class ContributionReward extends ExtendTruffleContract(SolidityContributionRewar
     const ethReward = web3.toBigNumber(options.ethReward);
     const externalTokenReward = web3.toBigNumber(options.externalTokenReward);
 
-    if ((nativeTokenReward < 0) || (reputationReward < 0) || (ethReward < 0) || (externalTokenReward < 0))
-    {
+    if (
+      nativeTokenReward < 0 ||
```
orenyodfat: Does this parameters can be type as unsigned somehow? so the is negative check can be avoided?

lib/contributionreward.js
```diff
@@ -149,7 +168,7 @@ class ContributionReward extends ExtendTruffleContract(SolidityContributionRewar
         reputationReward,
         ethReward,
         options.externalToken,
```
orenyodfat: option.externalToken should be removed .
It is passed as a separate param.

lib/contributionreward.js
```diff
@@ -158,11 +177,16 @@ class ContributionReward extends ExtendTruffleContract(SolidityContributionRewar
...
+      params.orgNativeTokenFee,
+      params.schemeNativeTokenFee,
+      params.voteParametersHash,
+      params.votingMachine
+    );
   }
 
   getDefaultPermissions(overrideValue) {
-    return overrideValue || '0x00000001';
+    return overrideValue || "0x00000001";
```
orenyodfat: It will be nice to have meaning full permission..
which can be defined somewhere in a general header file.
e.g REGISTERED = "0x00000001".
so it will looks like return overrideValue || REGISTERED;

by the way , please note that permission 0x00000000 is same as permission 0x00000001 as at the moment this scheme is register to the controller ..this bit is set automatically .

lib/globalconstraintregistrar.js
```diff
@@ -80,46 +88,50 @@ export class GlobalConstraintRegistrar extends ExtendTruffleContract(SolidityGlo
...
-    return await this._setParameters(params.voteParametersHash, params.votingMachine);
+    return await this._setParameters(
+      params.voteParametersHash,
+      params.votingMachine
+    );
   }
 
   getDefaultPermissions(overrideValue) {
-    return overrideValue || '0x00000007';
+    return overrideValue || "0x00000007";
```
orenyodfat: Why this scheme as default value different from contributionreward default value ?

lib/organization.js
```diff
@@ -86,9 +85,17 @@ export class Organization {
     org.reputation = await Reputation.at(reputationAddress);
 
     org.votingMachine = await AbsoluteVote.at(options.votingMachine);
```
orenyodfat: Do we need AbsoluteVote explicitly here.
The voting machine should be pass as a param for each scheme (if this scheme need voting machine as param).

lib/organization.js
```diff
@@ -113,7 +121,9 @@ export class Organization {
       initialSchemesParams.push(paramsHash);
       initialSchemesTokenAddresses.push(await scheme.nativeToken());
       initialSchemesFees.push(await scheme.fee());
-      initialSchemesPermissions.push(scheme.getDefaultPermissions(/* supply options.permissions here? */));
+      initialSchemesPermissions.push(
+        scheme.getDefaultPermissions(/* supply options.permissions here? */)
```
orenyodfat: why default ? It should be given as param to the function

lib/organization.js
```diff
@@ -123,19 +133,18 @@ export class Organization {
...
     // we must do this after setSchemes, because that one approves the transactions
     // (but that logic shoudl change)
-    for (let i=0; i < initialSchemesAddresses.length; i = i + 1) {
+    for (let i = 0; i < initialSchemesAddresses.length; i = i + 1) {
       const scheme = await SchemeRegistrar.at(initialSchemesAddresses[i]);
       const token = await DAOToken.at(initialSchemesTokenAddresses[i]);
-      const fee  = initialSchemesFees[i];
+      const fee = initialSchemesFees[i];
       await token.transfer(org.avatar.address, fee);
       await scheme.registerOrganization(org.avatar.address);
```
orenyodfat: This is not needed any more as the registerOrganization is done in setSchemes for the case it is  universalScheme.
Take a look at daostack https://github.com/daostack/daostack/blob/master/contracts/universalSchemes/GenesisScheme.sol#L88

lib/organization.js
```diff
@@ -123,19 +133,18 @@ export class Organization {
...
 
-
     // transfer what we need for fees to register the organization at the given schemes
     // TODO: check if we have the funds, if not, throw an exception
     // fee = await org.schemeRegistrar.fee())
     // we must do this after setSchemes, because that one approves the transactions
     // (but that logic shoudl change)
-    for (let i=0; i < initialSchemesAddresses.length; i = i + 1) {
+    for (let i = 0; i < initialSchemesAddresses.length; i = i + 1) {
       const scheme = await SchemeRegistrar.at(initialSchemesAddresses[i]);
```
orenyodfat: Why SchemeRegistrar ???

lib/utils.js
```diff
@@ -148,84 +154,96 @@ export function getDefaultAccount() {
...
+export const ExtendTruffleContract = superclass =>
+  class {
+    constructor(contract) {
+      this.contract = contract;
+      for (const i in this.contract) {
+        if (this[i] === undefined) {
+          this[i] = this.contract[i];
+        }
       }
+      // for (var prop in this.contract) {
```
orenyodfat: Please remove this commented code

lib/utils.js
```diff
@@ -148,84 +154,96 @@ export function getDefaultAccount() {
...
-  }
+    /**
+     * Call setParameters on this scheme.
+     * Returns promise of parameters hash.
+     * If there are any parameters, then this function must be overridden by the subclass to provide them.
+     * @param overrides -- object with properties whose names are expected by the scheme to correspond to parameters.  Overrides the defaults.
+     *
+     * Should have the following properties:
+     *
+     *  for all:
```
orenyodfat: Does this comment is still valid ?

lib/utils.js
```diff
@@ -148,84 +154,96 @@ export function getDefaultAccount() {
...
-  async setParams(params) {
-    return await '';
-  }
+    /**
+     * Call setParameters on this scheme.
+     * Returns promise of parameters hash.
+     * If there are any parameters, then this function must be overridden by the subclass to provide them.
+     * @param overrides -- object with properties whose names are expected by the scheme to correspond to parameters.  Overrides the defaults.
+     *
+     * Should have the following properties:
```
orenyodfat: > Should have the following properties:
 -   *		 +     *
 -   *  for all:		 +     *  for all:
 -   *    voteParametersHash		 +     *    voteParametersHash
 -   *    votingMachine -- address		 +     *    votingMachine -- address
 -   *		 +     *
 -   *  for ContributionReward:		 +     *  for ContributionReward:
 -   *    orgNativeTokenFee -- number		 +     *    orgNativeTokenFee -- number
 -   *    schemeNativeTokenFee -- number		 +     *    schemeNativeTokenFee -- number

Does this comment is a valid one ?

lib/utils.js
```diff
@@ -148,84 +154,96 @@ export function getDefaultAccount() {
...
+     *
+     *  for all:
+     *    voteParametersHash
+     *    votingMachine -- address
+     *
+     *  for ContributionReward:
+     *    orgNativeTokenFee -- number
+     *    schemeNativeTokenFee -- number
+     */
+    // eslint-disable-next-line no-unused-vars
```
orenyodfat: Why disable eslint here ?

package-scripts.js
```diff
@@ -27,122 +38,121 @@ module.exports = {
...
+        run: `testrpc -l 6500000 --account="0x0191ecbd1b150b8a3c27c27010ba51b45521689611e669109e034fd66ae69621,9999999999999999999999999999999999999999999" --account="0x00f360233e89c65970a41d4a85990ec6669526b2230e867c352130151453180d,9999999999999999999999999999999999999999999" --account="0x987a26abca7432016104ce2f24ce639340e25afe06ac69f68791399e7a5d1028,9999999999999999999999999999999999999999999" --account="0x89af34b1b7347834048b99423dad174a14bf14540d720d72c16ae92e94b987cb,9999999999999999999999999999999999999999999" --account="0xc867be647eb2bc51e4c0d61066859875cf3634fe949b6f5f85c69ab90e485b37,9999999999999999999999999999999999999999999" --account="0xefabcc2377dee5e51b5a9e65a3854aec85fbbec3cb584d8ad4f9679869fb33c6,9999999999999999999999999999999999999999999"`
```
orenyodfat: testrpc -> ganache-cli  
See https://github.com/daostack/daostack/blob/master/package.json#L7

package-scripts.js
```diff
@@ -27,122 +38,121 @@ module.exports = {
...
       },
       testrpcDb: {
```
orenyodfat: this need to be removed.

package-scripts/archiveTestrpcDb.js
```diff
@@ -1,9 +1,11 @@
-const archiver = require('archiver');
+const archiver = require("archiver");
```
orenyodfat: This file should be removed.

package-scripts/unArchiveTestrpcDb.js
```diff
@@ -1,5 +1,5 @@
-const decompress = require('decompress');
+const decompress = require("decompress");
```
orenyodfat: This file should be removed,

test/contributionreward.js
```diff
@@ -1,21 +1,25 @@
...
 
 export async function proposeContributionReward(org, accounts) {
   const schemeRegistrar = await org.scheme("SchemeRegistrar");
-  const ContributionReward = await org.scheme('ContributionReward');
+  const ContributionReward = await org.scheme("ContributionReward");
 
-  const votingMachineHash = await org.votingMachine.getParametersHash(org.reputation.address, 50, true);
+  const votingMachineHash = await org.votingMachine.getParametersHash(
+    org.reputation.address,
+    50,
```
orenyodfat: From this value come from ? Is it default ?

orenyodfat:
Though this PR has no code logic changes I take the opportunity to add some comments.

lib/arc.js
dkent600: This PR is about formatting, that is all

lib/contributionreward.js
dkent600: This PR is about formatting, that is all. I would prefer to take note of other comments and address them elsewhere.

lib/contributionreward.js
dkent600: This PR is about formatting, that is all. I would prefer to take note of other comments and address them elsewhere.

dkent600:
@orenyodfat It is great to have those comments, thanks!  We need to create separate issues for them.  Some of them are already part-and-parcel of other issues.  I can take care of creating new issues where they are needed.

lib/organization.js
dkent600: Good question!  That code is going away in any case.

orenyodfat:
Most of the comments refer to new code/comments change (+green).
I guess it is a good opportunity to address these before formatting the code..because it might need reformatting after address these issues,

dkent600:
I really don't want to do anything but formatting in this one.

lib/utils.js
dkent600: because there is an unused var, but that is intentional

.eslintignore
dkent600: I fixed this

lib/utils.js
dkent600: At this moment, yes, but as we add different schemes and voting machines, no

lib/arc.js
dkent600: Because of this code in utils.js:

```
    preWeb3 = new Web3(
      new Web3.providers.HttpProvider("http://localhost:8545")
    );
```

lib/organization.js
orenyodfat: I think there is minimum permission per scheme according to what it is doing.
e.g a SchemeRegistrar scheme should at least has registrar permission.(0x3)..and so on..
But if a user would like to create an org with a "special" scheme it should be able to give it a permission it needed.

lib/utils.js
orenyodfat: Lets remove that if it is not valid.

lib/utils.js
orenyodfat: how come unused var is intentional ?

dkent600: it is like an abstract method, never directly used

orenyodfat: so how about  : 
async setParams(params) {
    return await params;
  }
instead of : 
async setParams(params) {
    return await "";
  }
If its any way never directly used.

orenyodfat: I guess we should try to avoid linter exceptions if we can

dkent600: That suggestion would return the wrong datatype and thus be misleading to anyone trying to understand the code.

So I did it like this:

```
     */
    async setParams(params) {
      params; // avoid lint error
      return await "";
    }
```

orenyodfat: ok

PR #22, opened 2018-01-04
Clean up settings/configuration

tibetsprague:
Add configuration using nconf with default values living in config/default.json
Apps that use this module can override those settings with a daostack.json file

Addresses https://github.com/daostack/arc-js/issues/21 and 
https://github.com/daostack/arc-js/issues/3


lib/arc.d.ts
```diff
@@ -54,29 +56,24 @@ declare module "daostack-arc-js" {
...
-    address: string;
-    blockHash: string;
-    blockNumber: number;
-    data: string;
-    logIndex: number;
-    topics: Array<string>;
-    transactionHash: string;
-    transactionIndex: number;
-    type: string;
+    address: string,
```
dkent600: Typescript interfaces use semi-colons, not commas

lib/config.js
```diff
@@ -0,0 +1,14 @@
+import nconf from 'nconf';
+
+// Load configuration
+// First look for overrides from the command line or environment variables
+nconf.argv().env();
+
+// Then look for overrides from modules that are importing arc-js
+const path = 'daostack.json'; // Path to configuration override file in calling module
+nconf.file("external", path);
```
dkent600: would this be better named "arc-js.config"?

dkent600: To what location in the application does 'daostack.json' resolve?  How can the application change this location?

dkent600: What if the application wants to supply config values that are computed at runtime, rather hardcoded on disk?  How is that done?

lib/contracts.js
```diff
@@ -0,0 +1,83 @@
...
+    },
+    AbsoluteVote: {
+      contract: AbsoluteVote,
+      address: absoluteVote.address,
+    }
+  };
+
+  return {
+    allContracts : contracts,
+    defaultVotingMaching: contracts.AbsoluteVote,
```
dkent600: As noted previously, `defaultVotingMaching` is mispelled.

lib/organization.js
```diff
@@ -1,18 +1,17 @@
...
 const DAOToken = requireContract("DAOToken");
 const GenesisScheme = requireContract("GenesisScheme");
 const Reputation = requireContract("Reputation");
 // import { GlobalConstraintRegistrar } from   './globalconstraintregistrar.js';
 const AbsoluteVote = requireContract("AbsoluteVote");
 // import { UpgradeScheme } from   './upgradescheme.js';
-
-import { getSettings } from "./settings.js";
-import { SchemeRegistrar } from "./schemeregistrar.js";
+import { SchemeRegistrar } from './schemeregistrar.js';
```
dkent600: I just checked in a big code reformatting PR that converted all the single quotes to double quotes, according to rules of prettify.  But I neglected to incorporate the rule into eslint.  I'm afraid we're going to have a lot of thrashing and more painful diff's if we don't settle on a standard with respect to quotes.

lib/contracts.js
```diff
@@ -0,0 +1,83 @@
...
+  const upgradeScheme = await UpgradeScheme.deployed();
+  const absoluteVote = await AbsoluteVote.deployed();
+
+  /**
+   * `contract` here is an uninitialized instance of ExtendTruffleContract,
+   * basically the class factory.
+   * Calling contract.at() (a static method on the class) will return a
+   * the properly initialized instance of ExtendTruffleContract.
+   */
+  const contracts = {
```
dkent600: I'm really not a big fan of this dictionary style of storing the contracts, as an object whose property names are the names of contracts.  But at this moment I am at a loss for a better idea.  Maybe it is best.

lib/organization.js
```diff
@@ -167,9 +166,9 @@ export class Organization {
 
     // TODO: we now just set the default voting machine, and assume it is used
     // throughout, but this assumption is not warranted
-    const settings = await getSettings();
-    if (settings.votingMachine) {
-      org.votingMachine = AbsoluteVote.at(settings.votingMachine);
+    const contracts = await getDeployedContracts();
+    if (contracts.defaultVotingMachine) {
+      org.votingMachine = AbsoluteVote.at(contracts.defaultVotingMachine);
```
dkent600: I note that `defaultVotingMachine` is not misspelled here.

lib/wallet.js
```diff
@@ -1,27 +1,17 @@
...
-import { getWeb3 } from "./utils.js";
+import { config } from './config.js';
+import * as ethers from 'ethers';
+const bip39 = require('bip39');
+import  { getWeb3 } from './utils.js';
 
 import { Organization } from "./organization.js";
 
-// TODO: move all the provider connection stuff to settings.js
-const ethProvider = process.env.ETH_PROVIDER || "local"; // 'infura', 'etherscan', otherwise will default to local RPC (testrpc, Geth, or Parity..)
```
dkent600: Should these settings be coming from config.js?

package-scripts.js
```diff
@@ -11,6 +11,7 @@ const {
 const env = require("env-variable")();
 const joinPath = require("path.join");
 const cwd = require("cwd")();
+const config = require('./dist/config.js').config;
```
dkent600: What is dist/config.js going to contain at this point?  There is no application context here.

lib/arc.d.ts
tibetsprague: weird, i didnt touch that

lib/config.js
tibetsprague: im still really not sure what naming makes sense for anything. to me these parameters are more relevant to daostack than arc since they are more in reference to interaction with the blockchain.

tibetsprague: The location is the root of the project that imports that arc-js module. I was thinking of adding the ability pass in a different path to this file, just didnt get to it yet. let me think about the best way to support overriding config values at runtime. you can do it already since nconf makes it easy, you can just call nconf.overrides({ key: value}), but i will think about whether there a better interface for this

lib/contracts.js
tibetsprague: me either, this is just a start

lib/organization.js
tibetsprague: not a big fan of prettify so far, but ok

lib/wallet.js
tibetsprague: what do you mean? they are in this PR

package-scripts.js
tibetsprague: it loads the configuration data into the nconf object to be used in package-scripts.js

lib/organization.js
tibetsprague: fixe

package-scripts.js
dkent600: What if the user wants to change the configuration data, like the current network?  I don't think it is a good idea to require that users alter configuration files within a package installation folder -- the installation may change and then the config is overwritten.

lib/config.js
dkent600: The user shouldn't need to know that these settings pertain to arc, arc is encapsulated, and the settings may not *always* pertain to arc, they may also someday be specific to arc-js.

DAOstack is the name of the company, not the package.

lib/contracts.js
dkent600: With a fresh morning mind it occurred to me that this contracts object should perhaps share at least some of the structure of the data that we are going to provide about current and historical contracts/schemes.  They two sets of data should be compatible.

lib/wallet.js
dkent600: Sorry, I misread the '-' versus '+'.

lib/contracts.js
dkent600: With a fresh morning mind it occurred to me that this contracts object should perhaps share at least some of the structure of the data that we are going to provide about current and historical contracts/schemes. They two sets of data should be compatible in schema.

lib/config.js
tibetsprague: i'm still confused on naming, it seems weird to call both repos arc to me. and daostack is in the name of the package. we could split up the settings though into multiple sections though

lib/contracts.js
tibetsprague: totally agree

package-scripts.js
tibetsprague: they can add a daostack.json file as above, or use command line arguments or environment variables

lib/config.js
dkent600: My understanding has been that the contracts are "Arc" and the JavaScript library is  "Arc-Js", each with "DAOstack-" prepended.

It's also my expectation that all of that could still be in flux...

I agree that the repos should have the same names.

package-scripts.js
dkent600: Where would command line arguments be supplied?

Are you saying the user can use environment variables instead of, or in addition to, daostack.json?  Which would take precedence?

Can we be sure this is all documented somewhere?

Just to confirm:  Have you tested using all three methods, both when running in a package script and when running in a straight application (not in nodejs)?

dkent600: Where would command line arguments be supplied?
Are you saying the user can use environment variables instead of, or in addition to, daostack.json? Which would take precedence?
Can we be sure this is all documented somewhere?
Just to confirm: Have you tested using all three methods, both when running in a package script and when running in a straight application (not in nodejs)?

tibetsprague: command line arguments take the most precendence, then env variables, then the config file. i added more documentation to the README

lib/config.js
tibetsprague: i renamed the file to arc.config.json

dkent600: I think this should be arc-js.config.json.

lib/contracts.js
dkent600: Is the plan then to refactor this when we do the work on supplying the full historical contracts?

tibetsprague: yeah

lib/config.js
tibetsprague: hmm, so ugly. arc seems fine to me personally

dkent600: Yeah, the problem is that arc refers to a different package.  This one is named arc-js, and that is what we're configuring here.  Better to be ugly than to be confusing or misleading, IMHO.

tibetsprague: well sort of, i really think of this file as configuring both arc and arc-js, together. they work hand in hand. i could see arc.config.json being used to configure the arc repo too.

config/default.json
```diff
@@ -0,0 +1,5 @@
+{
+  "providerUrl": "http://localhost:8545",
+  "network": "kovan",
+  "gasLimit": 6900000
```
orenyodfat: 6500000 is enough .why increase ?

package-scripts.js
```diff
@@ -46,7 +47,7 @@ module.exports = {
...
+        run: `testrpc -l ${config.get('gasLimit')} --account="0x0191ecbd1b150b8a3c27c27010ba51b45521689611e669109e034fd66ae69621,9999999999999999999999999999999999999999999" --account="0x00f360233e89c65970a41d4a85990ec6669526b2230e867c352130151453180d,9999999999999999999999999999999999999999999" --account="0x987a26abca7432016104ce2f24ce639340e25afe06ac69f68791399e7a5d1028,9999999999999999999999999999999999999999999" --account="0x89af34b1b7347834048b99423dad174a14bf14540d720d72c16ae92e94b987cb,9999999999999999999999999999999999999999999" --account="0xc867be647eb2bc51e4c0d61066859875cf3634fe949b6f5f85c69ab90e485b37,9999999999999999999999999999999999999999999" --account="0xefabcc2377dee5e51b5a9e65a3854aec85fbbec3cb584d8ad4f9679869fb33c6,9999999999999999999999999999999999999999999"`
```
orenyodfat: use ganache-cli instead of testrpc (as daostack-arc is using).

orenyodfat:
do we need contracts folder here? the contracts json are from daostack-arc npm

orenyodfat:
SetScheme now as different interface so I am not sure how it will pass tests  : 
https://github.com/daostack/arc-js/blob/master/lib/organization.js#L130 ??

package-scripts.js
dkent600: using ganache-cli is included in what I'm doing in https://github.com/daostack/arc-js/issues/17

dkent600: Using ganache-cli is included in what I'm doing for https://github.com/daostack/arc-js/issues/17

dkent600:
>SetScheme now as different interface so I am not sure how it will pass tests :
https://github.com/daostack/arc-js/blob/daostack_config/lib/organization.js#L129 ??
> also registerOrganization is done within setScheme for universalScheme... so you can and should avoid that at lib/organization.js

These changes are all happening as part of https://github.com/daostack/arc-js/issues/17
But they are blocked by https://github.com/daostack/daostack/issues/226

leviadam:
All the deployment issues will be handled by arc-js.
I think we can close this PR.

tibetsprague:
@leviadam improving how we handle configuration settings in Arc JS is valuable no matter what, im not sure why you closed this

config/default.json
tibetsprague: i needed to at some point, but if 650000 is working right now thats fine

leviadam:
Hi @tibetsprague, sorry for that. It was a mistaken tab.
I actually read the PR in the daostack repo and meant to comment on that one. Got to this one instead.

PR #31, opened 2018-01-18
Add getters to Organization for name, token name and symbol

tibetsprague:
Update and fix typescript definitions

Bump npm version


lib/arc.d.ts
```diff
@@ -507,5 +511,9 @@ declare module "daostack-arc-js" {
       opts: ProposeContributionParams
     ): Promise<TransactionReceiptTruffle>;
     setParams(params: ContributionRewardParams): Promise<string>;
+
```
dkent600: I'm not sure these events should be part of this PR, better to be part of a broader task aimed at adding better support for events across all of the wrappers.  In any case, I don't see that they've been implemented in the ContributionReward class.

tsconfig.json
```diff
@@ -2,7 +2,6 @@
   "compilerOptions": {
     "module": "commonjs",
     "noImplicitAny": true,
-    "outFile": "./build/tsc.js",
```
dkent600: What is this for?

lib/arc.d.ts
```diff
@@ -234,6 +234,10 @@ declare module "daostack-arc-js" {
     // vote(proposalId, choice, params);
     static new(options: OrganizationNewConfig): Promise<Organization>;
     static at(avatarAddress: string): Promise<Organization>;
+
+    getName(): string;
+    getTokenName(): string;
```
dkent600: Can you please add comments describing what these three new methods do? (like *what* name is returned, and so on)?  Thanks.

tsconfig.json
tibetsprague: I couldn't run tsc to test the syntax of the definitions file without removing this

lib/arc.d.ts
tibetsprague: thats true, and this is a good opportunity to do that work, but I needed these definitions to get my old alchemy demo working again. these functions come directly from the truffle contract

lib/arc.d.ts
tibetsprague: sure

lib/arc.d.ts
tibetsprague: are you ok with leaving these in for now while we work out how we want to handle the event functionality in the wrappers? in fact we may decide to just leave as is and pass through to the truffle contract handling for now

dkent600: sure

lib/arc.d.ts
dkent600: Heh, should be "Organization name", not "DAO"  Sorry ;-)

lib/organization.js
```diff
@@ -274,14 +274,26 @@ export class Organization {
     return this.votingMachine.vote(proposalId, choice, params);
   }
 
+  /**
```
dkent600: Organization, not DAO here and below too

Issue #30, opened 2018-01-18
Remove migrateContracts.initialize script


dkent600:
Remove migrateContracts.initialize script, it is no longer needed.

dkent600:
![](https://github.trello.services/images/mini-trello-icon.png) [Demo:  global constraints dashboard should list and work with gc's registered with the controller](https://trello.com/c/q6bd3gIF/24-demo-global-constraints-dashboard-should-list-and-work-with-gcs-registered-with-the-controller)

Issue #28, opened 2018-01-17
add getGlobalConstraints function to Organization 


dkent600:
Just as the Organization class has a `schemes` method that returns all the schemes registered into the organization, it needs a `globalConstraints` method that returns all the global constraints registered into it.

This is needed to accompany:  https://github.com/daostack/Alchemy-Aurelia/issues/11

dkent600:
![](https://github.trello.services/images/mini-trello-icon.png) [Demo:  global constraints dashboard should list and work with gc's registered with the controller](https://trello.com/c/q6bd3gIF/24-demo-global-constraints-dashboard-should-list-and-work-with-gcs-registered-with-the-controller)

PR #33, opened 2018-01-19
added Organization.globalConstraints

dkent600:
This addresses #28 - add getGlobalConstraints function to Organization
and #30 - Remove migrateContracts.initialize script

It is required to be able to proceed with a PR for https://github.com/daostack/Alchemy-Aurelia/issues/11


lib/organization.js
```diff
@@ -145,24 +145,22 @@ export class Organization {
...
-      });
+      return schemes.filter(s => s.name === name);
     } else {
       return schemes;
     }
   }
 
   /**
-   * returns schemes currently in this Organization as Array<OrganizationSchemeInfo>
+   * returns schemes currently in this DAO as Array<OrganizationSchemeInfo>
```
tibetsprague: why this change to DAO? seems confusing to use both, we should pick one or the other

lib/organization.js
```diff
@@ -273,4 +271,97 @@ export class Organization {
...
+    const constraints = await this._getConstraints();
+    if (name) {
+      return constraints.filter(s => s.name === name);
+    } else {
+      return constraints;
+    }
+  }
+
+  /**
+   * returns global constraints currently in this Organization, as OrganizationGlobalConstraintInfo
```
tibetsprague: Organization or DAO?

lib/organization.js
```diff
@@ -273,4 +271,97 @@ export class Organization {
...
+   */
+  async _getConstraints() {
+    // private method returns all registered schemes.
```
tibetsprague: constraints not schemes, also why do we need this comment alongside the block comment above the function?

lib/organization.js
```diff
@@ -273,4 +271,97 @@ export class Organization {
...
+      const paramsHash = eventsArray[i].args._params;
+
+      const info = {
+        address: address,
+        paramsHash: paramsHash,
+        // will be undefined if not a known scheme
+        name: arcTypesMap.get(address)
+      };
+
+      constraintsMap.set(address, info);
```
tibetsprague: why not just check if its registered here instead of adding all of them to the Map and then iterating through the whole thing again to see which ones are registered?

test/organization.js
```diff
@@ -76,4 +77,67 @@ describe("Organization", () => {
 
     assert.equal((await organization.schemes()).length, 4);
   });
+
+  it("has a working globalConstraints() function to access its schemes", async () => {
```
tibetsprague: confusing to use the word schemes here

lib/organization.js
dkent600: yeah, there are a lot of old comments in this code from long before my time that I tend to just block out of my mind...

lib/organization.js
dkent600: the map eliminates dups

lib/organization.js
dkent600: agreed, done

lib/organization.js
dkent600: done

lib/organization.js
dkent600: deleted

lib/organization.js
dkent600: commented

test/organization.js
dkent600: done

lib/organization.js
tibetsprague: so you think its more likely that there will be dupes to eliminate first rather than unregistered constraints to eliminate first? either way its not a big deal

Issue #24, opened 2018-01-11
Catch up with new Arc schemes and other entities


dkent600:
We need to survey Arc to find out what new schemes and other entities need to be pulled in, for example,  but not limited-to, new voting machines.

tibetsprague:
we need to do a survey to find out what functionality is missing

tibetsprague:
![](https://github.trello.services/images/mini-trello-icon.png) [Catch up ArcJS wrappers to match all contracts and necessary functionality in Arc](https://trello.com/c/QknArAiV/33-catch-up-arcjs-wrappers-to-match-all-contracts-and-necessary-functionality-in-arc)

dkent600:
Some that do not at this writing have wrappers:

GenesisScheme
OrganizationRegister
SimpleICO
VestingScheme
VoteInOrganizationScheme
EmergentVoteScheme
QuorumVote

dkent600:
What we will do for this:

* GenesisScheme
* VestingScheme
* VoteInOrganizationScheme
* Governance

Others can be requested in future tickets.

dkent600:
Done

PR #57, opened 2018-01-31
Add GenesisScheme wrapper

dkent600:
This addresses part of https://github.com/daostack/Arc.js/issues/24

### DIFF
--- a/lib/arc.d.ts
+++ b/lib/arc.d.ts
@@ -65,7 +65,6 @@ declare module "daostack-arc.js" {
    */
   export interface ArcDeployedContracts {
     allContracts: ArcDeployedContractNames;
-    defaultVotingMachine: ArcContractInfo;
 
     /**
      * All deployed schemes
@@ -118,9 +117,9 @@ declare module "daostack-arc.js" {
   }
 
 
-   /********************************
-   * ExtendTruffleContract.js
-   */
+  /********************************
+  * ExtendTruffleContract.js
+  */
   export interface TransactionLog {
     address: string;
     blockHash: string;
@@ -207,9 +206,20 @@ declare module "daostack-arc.js" {
   }
 
   export interface FounderConfig {
+    /**
+     * Founders' address
+     */
     address: string;
-    tokens: string | number; // in Wei
-    reputation: string | number;
+    /**
+     * string | BigNumber array of token amounts to be awarded to each founder.
+     * Should be given in Wei.
+     */
+    tokens: string | BigNumber.BigNumber;
+    /**
+     * string | BigNumber array of reputation amounts to be awarded to each founder.
+     * Should be given in Wei.
+     */
+    reputation: string | BigNumber.BigNumber;
   }
 
   export interface NewDaoVotingMachineConfig {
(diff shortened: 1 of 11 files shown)


lib/arc.d.ts
```diff
@@ -679,4 +699,42 @@ declare module "daostack-arc.js" {
...
+     */
+    getAvatar(): any;
+    /**
+     * returns the address of the controller
+     */
+    getControllerAddress(): string;
+    /**
+     * Returns a TruffleContract for the controller.  Could be
+     * either UController or Controller.  You can know which one
+     * by checking the AvatarSerrvice instance property `isUController`.
```
tibetsprague: spelling

lib/avatarService.js
```diff
@@ -0,0 +1,106 @@
...
+      const avatar = await this.getAvatar();
+      this._controllerAddress = await avatar.owner();
+    }
+    return this._controllerAddress;
+  }
+
+  /**
+   * Returns a TruffleContract for the controller.  Could be
+   * either UController or Controller.  You can know which one
+   * by checking the AvatarSerrvice instance property `isUController`.
```
tibetsprague: spelling

lib/avatarService.js
```diff
@@ -0,0 +1,106 @@
...
+      const avatar = await this.getAvatar();
+      this._nativeReputationAddress = await avatar.nativeReputation();
+    }
+    return this._nativeReputationAddress;
+  }
+
+  /**
+   * Returns the avatar's native reputation TruffleContract.
+   */
+  async getNativeReputation() {
```
tibetsprague: do we need to call it "Native"? seems like reputation is always native to a DAO

lib/contracts/genesisscheme.js
```diff
@@ -0,0 +1,186 @@
+"use strict";
+import { Utils } from "../utils";
```
tibetsprague: I like to order imports alphabetically

lib/contracts/genesisscheme.js
```diff
@@ -0,0 +1,186 @@
...
+const Avatar = Utils.requireContract("Avatar");
+import { AbsoluteVote } from "./absoluteVote.js";
+
+const dopts = require("default-options");
```
tibetsprague: and put external libraries at the top in a separate section

lib/contracts/genesisscheme.js
```diff
@@ -0,0 +1,186 @@
...
+      name: undefined,
+      tokenName: undefined,
+      tokenSymbol: undefined,
+      founders: [],
+      universalController: true
+    };
+
+    const options = dopts(opts, defaults, { allowUnknown: true });
+
+    if (!options.name) {
```
tibetsprague: this works with undefined? I thought you had to check typeof === "undefined"?

lib/contracts/genesisscheme.js
```diff
@@ -0,0 +1,186 @@
...
+    if (!options.tokenName) {
+      throw new Error("DAO token name is not defined");
+    }
+
+    if (!options.tokenSymbol) {
+      throw new Error("DAO token symbol is not defined");
+    }
+
+    const web3 = Utils.getWeb3();
+    const controllerAddress = options.universalController ? (await (await Utils.requireContract("UController")).deployed()).address : Utils.NULL_ADDRESS;
```
tibetsprague: awkward line, can we break it up?

tsconfig.json
```diff
@@ -15,7 +17,6 @@
     "dist",
   ],
   "typeRoots": [
-    "node_modules/@types",
-    "node_modules/web3-typescript-typings"
```
tibetsprague: this is no longer needed? I think it might be, when this package is used by another package

lib/avatarService.js
dkent600: it correlates to the contract property name, so I would defer to that unless the contract gurus suggest otherwise.

lib/contracts/genesisscheme.js
dkent600: dopts handles the undefined check, but it has a weakness that it considers "null" to be a valid value of a required property (because null isn't undefined).  So we do a manual check.

In any case, this would also catch undefined.

tsconfig.json
dkent600: Vanille doesn't seem to need it.

dkent600: BTW, I did a thorough check this time that arc.d.ts compiles.

dkent600: (so does the javascript!)

PR #60, opened 2018-02-01
Add vesting scheme wrapper

dkent600:
Continues to address #24 by adding the VestingScheme wrapper.

Also cleans up the d.ts a little, and adds some stuff that got missed when creating the GenesisScheme wrapper.

### DIFF
--- a/lib/contracts/vestingscheme.js
+++ b/lib/contracts/vestingscheme.js
@@ -1,7 +1,7 @@
 "use strict";
 const dopts = require("default-options");
 
-import { ExtendTruffleContract, ArcTransactionResult } from "../ExtendTruffleContract";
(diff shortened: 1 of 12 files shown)


lib/contracts/vestingscheme.js
```diff
@@ -0,0 +1,209 @@
...
+  }
+
+
+  /**
+   * see CreateVestingAgreementConfig
+   */
+  _defaultCreateOptions = {
+    beneficiary: undefined,
+    returnOnCancelAddress: undefined,
+    startingBlock: null,
```
tibetsprague: why is this null and the rest undefined?

dkent600: when dopts sees undefined it throws an exception if the value comes in undefined.  any other value becomes the default.

gotta say, i'm not a big fan of dopts.  it is what was there when i came in.

PR #68, opened 2018-02-05
upgrade to arc v28

dkent600:
GenesisScheme => DaoCreator
remove "Log" from events
_globalconstraint => _globalConstraint in AddGlobalConstraint

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1,6 +1,6 @@
 "use strict";
 import { AvatarService } from "./avatarService";
-import { GenesisScheme } from "./contracts/genesisscheme";
+import { DaoCreator } from "./contracts/daocreator";
 import { getDeployedContracts } from "./contracts.js";
 import { Utils } from "./utils";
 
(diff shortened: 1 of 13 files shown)


lib/dao.js
```diff
@@ -288,12 +288,12 @@ export class DAO {
   /**
    * Returns promise of the DAOstack (Genesis) avatar address, or undefined if not found
    */
-  static async getDAOstack(genesisSchemeAddress) {
+  static async getDaoCreator(daoCreatorAddress) {
```
orenyodfat: @dkent600 What this function is doing ? Are you sure it is has the right name ?

migrations/2_deploy_organization.js
```diff
@@ -44,14 +44,14 @@ let schemeGCRegisterParams;
...
 
 module.exports = async function (deployer) {
-  // Deploy GenesisScheme:
+  // Deploy DaoCreator:
   // apparently we must wrap the first deploy call in a then to avoid
   // what seem to be race conditions during deployment
-  // await deployer.deploy(GenesisScheme)
-  deployer.deploy(GenesisScheme, { gas: 6015000 }).then(async () => {
-    genesisSchemeInst = await GenesisScheme.deployed();
+  // await deployer.deploy(DaoCreator)
```
orenyodfat: Please remove commented code

orenyodfat:
Is there a test dedicated to DaoCreator ?

dkent600:
@orenyodfat 

> Is there a test dedicated to DaoCreator ?

I didn't actually create any because it is getting used every time we call `DAO.new` in the tests so I figured at this moment that was good enough.  Maybe I was being lazy....let me know.

lib/dao.js
orenyodfat: @dkent600 What this function is doing ?

orenyodfat:
> I didn't actually create any because it is getting used every time we call DAO.new in the tests so I figured at this moment that was good enough. Maybe I was being lazy....let me know.

I guess it is good to have a dedicated standalone (as possible) test for each lib module.
I think that we should strive that the automated tests should cover each and every line of code in the lib.

PR #106, opened 2018-02-27
ContributionReward enhancements

dkent600:
Upgraded to Arc v34.

Cleaned up jslint packaging and config

Added functions to ContributionReward:

- redeemExternalToken
- redeemReputation
- redeemNativeToken
- redeemEther
- getDaoProposals
- getBeneficiaryRewards

### DIFF
--- a/lib/contracts/contributionreward.ts
+++ b/lib/contracts/contributionreward.ts
@@ -345,11 +345,19 @@ export class ContributionRewardWrapper extends ExtendTruffleContract {
     for (const proposal of proposals) {
 
       rewardsArray.push({
-        ethReward: await this.computeRemainingReward(proposal, "ethReward", options.avatar, RewardType.Eth),
-        externalTokenReward: await this.computeRemainingReward(proposal, "externalTokenReward", options.avatar, RewardType.ExternalToken),
-        nativeTokenReward: await this.computeRemainingReward(proposal, "nativeTokenReward", options.avatar, RewardType.NativeToken),
+        ethReward: await this.computeRemainingReward(
+          proposal, "ethReward", options.avatar, RewardType.Eth),
+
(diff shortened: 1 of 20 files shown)


lib/contracts/contributionreward.ts
```diff
@@ -157,6 +157,213 @@ export class ContributionRewardWrapper extends ExtendTruffleContract {
...
+    const tx = await this.contract.redeemExternalToken(
+      options.proposalId,
+      options.avatar,
+    );
+
+    return new ArcTransactionResult(tx);
+  }
+
+  /**
+   * Redeem reputation reward for the beneficiary of the proposal
```
tibetsprague: oh interesting, so anyone can do the redemption and decide to pay the gas for it?

lib/contracts/contributionreward.ts
```diff
@@ -157,6 +157,213 @@ export class ContributionRewardWrapper extends ExtendTruffleContract {
...
+
+    return proposals;
+  }
+
+  /**
+   * Return a list of executed (passed) proposals that have rewards
+   * waiting to be redeemed by the given beneficiary.
+   * @param opts
+   */
+  public async getBeneficiaryRewards(opts = {}): Promise<Array<ProposalRewards>> {
```
tibetsprague: maybe should say "unredeemed" in the name?

lib/contracts/contributionreward.ts
dkent600: Apparently.  I remarked (to myself) about that too.  Same for the other reward types.

lib/contracts/contributionreward.ts
dkent600: OK, I took the liberty of keeping the name but having it also return the total amounts redeemable.

PR #122, opened 2018-03-05
Add normalTxGasLimit

orenyodfat:
No need to use the overall gas limit (6300000) for all txs.
Use normalTxGasLimit for most txs

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -2,7 +2,7 @@
   "providerUrl": "http://127.0.0.1:8545",
   "network": "ganache",
   "gasLimit": 6300000,
-  "normalTxGasLimit": "4543760",
+  "gasLimit.runtime": "4543760",
   "apiToken": "",
   "defaultVotingMachine": "AbsoluteVote"
 }
--- a/config/default.json
+++ b/config/default.json
@@ -2,6 +2,7 @@
   "providerUrl": "http://127.0.0.1:8545",
   "network": "ganache",
   "gasLimit": 6300000,
+  "normalTxGasLimit": "4543760",
   "apiToken": "",
   "defaultVotingMachine": "AbsoluteVote"
 }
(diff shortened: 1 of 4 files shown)


dkent600:
@orenyodfat On what OS are you running this command?

(When I run it in a linux Bash shell on my machine it works fine.)

config/default.json
```diff
@@ -2,6 +2,7 @@
   "providerUrl": "http://127.0.0.1:8545",
   "network": "ganache",
   "gasLimit": 6300000,
```
dkent600: Please rename "gasLimit" to "gasLimit.deployment"

config/default.json
```diff
@@ -2,6 +2,7 @@
...
+  "normalTxGasLimit": "4543760",
```
dkent600: Please rename "normalTxGasLimit" to "gasLimit.runtime"

config/default.json
orenyodfat: done at 3f39eb4

orenyodfat:
> @orenyodfat Have you tested these changes when running an app with Metamask against both ganache and kovan, including forging DAOs and doing other stuff?

No. do you have a test procedure for that ?

dkent600:
@orenyodfat 

> No. do you have a test procedure for that ?

Nothing automated.  Right now we have to manually try it with Vanille or Alchemy and see what happens.

I think we need to do that to make sure we're not making things worse with these changes.  I will go ahead and do that since I think it will be a lot easier for me to do it.

orenyodfat:
@dkent600 👍

dkent600:
When with this PR:  https://github.com/daostack/arc.js/pull/127.

Created https://github.com/daostack/arc.js/issues/128 to follow up on forgeOrg

PR #123, opened 2018-03-06
Remove runAsync command.

orenyodfat:
-runAsync command (npm start test.ganache.runAsync)  is not working as the cmd
- it is running an empty cmd 

>node node_modules/run-with-ganache/bin/run-with-ganache --ganache-cmd "${ganacheCommand}" " "`  

- till it will be fixed i guess it should be removed.

-not sure why it is needed in first place.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ Arc.js runs against an Ethereum network where it assumes that the Arc contracts 
 
 To deploy contracts to a Ganache testnet, run the following scripts:
 
-Synchronously in a separate shell window:
+In a separate shell window:
 ```script
 npm explore @daostack/arc.js -- npm start test.ganache.run
 ```
--- a/README.md
+++ b/README.md
@@ -79,14 +79,9 @@ Arc.js runs against an Ethereum network where it assumes that the Arc contracts 
 
 To deploy contracts to a Ganache testnet, run the following scripts:
 
-Either synchronously in a separate shell window:
+Synchronously in a separate shell window:
 ```script
 npm explore @daostack/arc.js -- npm start test.ganache.run
-```
-
-Or asynchronously in the background of your current shell window:
-```script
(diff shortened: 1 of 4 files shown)


dkent600:
@orenyodfat On what OS are you running this command?

(When I run it in a linux Bash shell on my machine it works fine.)

orenyodfat:
> When I run it in a linux Bash shell on my machine it works fine.

How it is work if the cmd is empty ?
What the propose of that ?

When I run it in a linux Bash shell on my machine it works fine.

dkent600:
@orenyodfat 

Can you please change [this line](https://github.com/daostack/arc.js/blob/3763ae8c1274d370c0f0576484fa9fe7c6005c5b/package-scripts.js#L83) as follows, try again and report back:

```
runAsync: `node node_modules/run-with-ganache/bin/run-with-ganache --ganache-cmd "${ganacheCommand}" "echo"`
```

dkent600:
@orenyodfat 

Can you please let me know the OS you are using?  And try the change I proposed above?

> How it is work if the cmd is empty ? What the propose of that ?

The script requires _something_ there.  The empty command just makes it clean and works every place I've tried.

orenyodfat:
@dkent600 the purpose of run-with-ganache is to run a cmd in parallel with ganache .
if you are not running any special command in parallel with ganache the run_with_ganache should not be used.

README.md
```diff
@@ -79,16 +79,11 @@ Arc.js runs against an Ethereum network where it assumes that the Arc contracts
 
 To deploy contracts to a Ganache testnet, run the following scripts:
 
-Either synchronously in a separate shell window:
```
dkent600: Please change to:  

> In a separate shell window:

(Remove "synchronously")

dkent600:
@orenyodfat OK, you've convinced me to remove this command...

Issue #130, opened 2018-03-08
do not load solidity contracts at module load time


dkent600:
Currently solidity contracts are loaded at js module load time via calls to `Utils.requireContract`.

When running against Metamask, a problem occurs when MM is not yet unlocked when  `Utils.requireContract` is called -- `Utils.requireContract` returns `undefined` for the solidity contract, and thus, among other issues, arc.js wrappers are unable to instantiate themselves once the application starts use them.

We need to find a way to defer calling `Utils.requireContract`.

PR #132, opened 2018-03-08
Defer loading SolidityContract

dkent600:
Addresses: 

- https://github.com/daostack/arc.js/issues/130
- https://github.com/daostack/arc.js/issues/133

Defer loading SolidityContract
Renamed getScheme to getContractWrapper
Bumped version number
Pull founders from json file at deployment

**Breaking changes**: 
- Renamed getScheme to getContractWrapper
- Removed UController from `Contracts.getDeployedContracts`  (I really need to rename `Contracts.getDeployedContracts`)


lib/utils.ts
```diff
@@ -42,12 +42,21 @@ export class Utils {
...
    */
-  public static getWeb3(): any {
-    if (Utils.web3) {
+  public static getWeb3(forceReload: boolean = false): any {
+
+    if (forceReload) {
+      delete Utils.web3;
+      Utils.alreadyTriedAndFailed = false;
+      if (typeof window !== "undefined") {
+        delete window.web3;
```
tibetsprague: Hmm. Won’t this eliminate access to meta mask version of web3? I think we can reload web3 a different way. Maybe by calling get accounts?

dkent600: We set window.web3 below.

dkent600: Having said  that, I do question how it would affect the following ine:

`if (typeof web3 !== "undefined")`

dkent600: I'm not clear how "calling get accounts" would work.  Can you describe?

migrations/2_deploy_organization.js
```diff
@@ -30,50 +31,48 @@ module.exports = async (deployer) => {
...
-  const initRepInWei = [web3.toWei(10)];
-  const initTokenInWei = [web3.toWei(1000)];
   const orgNativeTokenFee = 0;
   const defaultVotingMachineParams = {
     preBoostedVoteRequiredPercentage: 50,
-    preBoostedVotePeriodLimit: 60,
-    boostedVotePeriodLimit: 60,
-    thresholdConstA: 1,
-    thresholdConstB: 1,
+    preBoostedVotePeriodLimit: 5184000, // 2 months
```
orenyodfat: from where these params come from ?

migrations/2_deploy_organization.js
```diff
@@ -30,50 +31,48 @@ module.exports = async (deployer) => {
...
+    boostedVotePeriodLimit: 604800, // 1 week
+    thresholdConstA: 1, // TODO: how many proposals can be boosted at a time. want about 5, have to read arc to figure out
```
orenyodfat: How this param is related to "how many proposals can be boosted at a time"?

migrations/2_deploy_organization.js
```diff
@@ -30,50 +31,48 @@ module.exports = async (deployer) => {
...
+    thresholdConstB: 1, // TODO: exponent of how hard to get the next one
     minimumStakingFee: 0,
-    quietEndingPeriod: 0,
-    proposingRepRewardConstA: 1,
-    proposingRepRewardConstB: 1,
-    stakerFeeRatioForVoters: 1,
-    votersReputationLossRatio: 10,
-    votersGainRepRatioFromLostRep: 80,
+    quietEndingPeriod: 7200, // Two hours
+    proposingRepRewardConstA: 5, // baseline rep rewarded TODO: good for now but Adam going to look up what this should be
```
orenyodfat: I would remove personal names .
It should be 

> baseline rep rewarded TODO: figure out the real default value.

migrations/2_deploy_organization.js
```diff
@@ -30,50 +31,48 @@ module.exports = async (deployer) => {
...
       tokenSymbol,
-      founders,
-      initTokenInWei,
-      initRepInWei,
-      // use the universal controller
-      0);
+      founders.map((f) => f.address),
+      founders.map((f) => web3.toWei(f.tokens)),
+      founders.map((f) => web3.toWei(f.reputation)),
+      // don't use the universal controller
```
orenyodfat: will change that to : 
> use single controller .

migrations/founders.kovan.json
```diff
@@ -0,0 +1,14 @@
+{
+  "founders": [
```
orenyodfat: please add a comment that this is for testing purpose only .

orenyodfat: why do we need to forge with founders for kovan ?Is that only for test ? 
Is that for alchemy on Kovan ? If yes, does some one has the prv keys for these  addresses ?

migrations/founders.kovan.json
```diff
@@ -0,0 +1,14 @@
+{
```
orenyodfat: do we need kovan as part of the file name ?

migrations/2_deploy_organization.js
```diff
@@ -30,50 +31,48 @@ module.exports = async (deployer) => {
   const orgName = "Genesis";
```
orenyodfat: Will change the name to "Genesis-Test" or something which defer that from "Genesis"

lib/utils.ts
```diff
@@ -42,9 +42,10 @@ export class Utils {
...
 
   /**
-   * Returns the web2 object.
+   * Returns the web3 object.
    * When called for the first time, web3 is initialized from the Arc.js configuration.
-   * Throws an exception when web3 cannot be initialized or there is no default client
+   * Throws an exception when web3 cannot be initialized.
+   * @param {boolean} forceReload true to reload/retry web3
    */
   public static getWeb3(): any {
```
orenyodfat: why this function is needed ?

migrations/2_deploy_organization.js
dkent600: They come from Adam and Tiber, as far as I know.  They are not complete.

migrations/founders.ganache.json
```diff
@@ -0,0 +1,19 @@
+{
```
orenyodfat: why need this file ?
why need ganache in the file name ?
why not just have founders.json for both ganache and kovan ?

migrations/2_deploy_organization.js
dkent600: @leviadam @tibetsprague

migrations/2_deploy_organization.js
dkent600: these are not complete

migrations/2_deploy_organization.js
orenyodfat: @leviadam ?

migrations/founders.kovan.json
dkent600: there is also one with ganache in the name

orenyodfat: why not have just founders.json (for both ganache and kovan ).

orenyodfat: for all networks

migrations/2_deploy_organization.js
tibetsprague: yup, Adam gave them to me, still gotta figure out the threshold params

migrations/founders.kovan.json
tibetsprague: this is why I dont think this file should be in the repo, we should be able to pass in a founders file if we want

tibetsprague: yes this is to setup the genesis DAO how we want it setup on deployment

migrations/2_deploy_organization.js
tibetsprague: ? this is the real genesis DAO, especially when we deploy to main net

migrations/2_deploy_organization.js
tibetsprague: * @dev threshold return the organization's score threshold which required by
     * a proposal to shift to boosted state.
     * This threshold is dynamically set and it depend on the number of boosted proposal.

tibetsprague: ok so @fmatan suggests thresholdConstA: 2, thresholdConstB: 10, to start

migrations/founders.kovan.json
orenyodfat: I agree with @tibetsprague  I feel that the genesisParams + the genesis_test founders + genesis dao should not be part of the arc.js .there should be passed as params via alchemy ...(manually or via file).

@tibetsprague I guess that this setup is not for the real final genesis DAO . it is just for demo. 
and it should be comment and maybe also should be reflected in the dao name.

tibetsprague: hmm, yeah I guess we could deploy without any DAOs and setup the Genesis DAO manually. but its nice to have it setup for us too for dev and testing

orenyodfat: "but its nice to have it setup for us too for dev and testing" 
Yes , but you can have it by default on alchemy (not arc.js)..no ?

tibetsprague: how so? deployment happens in Arc.js. I mean its nice having a DAO setup on deployment, so we dont have to do it manually each time.

migrations/founders.ganache.json
dkent600: merged both into founders.json

migrations/2_deploy_organization.js
dkent600: @orenyodfat can you please address Tibet's question, thanks!

migrations/founders.json
```diff
@@ -1,13 +1,33 @@
 {
   "founders": {
     "ganache": [
+      {
```
orenyodfat: why you need to mention the  'ganache' ?
founders are founders are founders .

migrations/2_deploy_organization.js
```diff
@@ -6,9 +6,8 @@ module.exports = async (deployer) => {
 
   const gasAmount = Config.get("gasLimit_deployment");
   const network = Config.get("network");
-  console.log(`Deploying to ${network}, gasLimit: ${gasAmount}`);
-
   const founders = require("./founders.json").founders[network];
+  console.log(`Deploying to ${network}, gasLimit: ${gasAmount},  ${founders.length} founders`);
```
orenyodfat: please remove console.log

orenyodfat: how does it pass the eslint ?

migrations/founders.json
orenyodfat: @dkent600 ?

migrations/2_deploy_organization.js
dkent600: this is a script that logs lots of information on the console as a normal and useful part of its function.  I would prefer to leave this in.  Does that work for you?

migrations/founders.json
```diff
@@ -0,0 +1,83 @@
+{
+  "founders": {
+    "ganache": [
+      {
+        "address": "0x8d4408014d165ec69d8cc9f091d8f4578ac5564f376f21887e98a6d33a6e3549",
```
orenyodfat: Why these adrresses are longer that kovan addresses ?
You can use the same address for kovan and ganache..no ?

migrations/founders.json
dkent600: This file contains both ganache and kovan founder data.

orenyodfat: why ?

migrations/founders.json
dkent600: these are private keys that ganache seems to like.  they correspond to the accounts that we tell ganache to create when we fire it up.

dkent600: The kovan addresses correspond to accounts on kovan that won't exist on ganache.

orenyodfat: - please replace private key with address .
- if it is not possible  change : "address" tp "private key" or at least comment that.

migrations/2_deploy_organization.js
orenyodfat: no.how the linter did not catch that ?

migrations/founders.json
tibetsprague: yeah these should all be public keys, on ganache too. the private keys won't work

dkent600: changed to addresses

migrations/2_deploy_organization.js
dkent600: Going with Genesis for the demo tomorrow

Issue #134, opened 2018-03-08
Why arc.js wrap truffle 'new','deploy' and 'at' functionality?


orenyodfat:
https://github.com/daostack/arc.js/blob/master/lib/ExtendTruffleContract.ts

Is that really needed ?

dkent600:
Short answer is that, together with `ContractWrapperFactory`, this gives us an architecture that supports the static `at`, `new`, and `deployed` methods while simultaneously giving us TypeScript typings for the handler classes and a good, straightforward class hierarchy.

orenyodfat:
It seems like it is adding unneeded layer of abstraction over truffle which makes the code library code none standard. 
Why set "setParams" and "getDefaultPermissions" are parts of ExtendTruffleContract (along with at, new and deploy) ?  what is the relation of that ?

dkent600:
![](https://github.trello.services/images/mini-trello-icon.png) [Rename ExtendTruffleContract to ContractWrapperBase](https://trello.com/c/qY0Fsqmg/104-rename-extendtrufflecontract-to-contractwrapperbase)

dkent600:
https://github.com/daostack/arc.js/commit/6589f5118324b5b2bf61c8f9e91007c31001902a

dkent600:
Quite a bit has changed this this issue was created, such as:

- `ExtendTruffleContract` has become `ContractWrapperBase`,which like its predecessor, provides wrapper functionality common to all wrapper classes.
- `at`, `new`, and `deployed` in `ContractWrapperBase` are renamed to `hydrateFromNew`, `hydrateFromAt`, `hydrateFromDeployed` to be more clear they are not creating new instances of the wrappers.
- All wrappers have a `factory` property which, true to its name, is the factory of the wrapper that provides another way to access `at`, `new` and `deployed` functions for wrapper classes.

@orenyodfat  Do you have further questions, or are we ready to close this?

dkent600:
No activity

Issue #16, opened 2017-12-30
write diagnostic info to console


dkent600:
Write diagnostic info to console, namely, the parameter values each time a contract method is invoked.

This would be off by default, turned on by a parameter in Arc.js's global configuration

dkent600:
![](https://github.trello.services/images/mini-trello-icon.png) [Add logging and log levels](https://trello.com/c/RJW14vFq/19-add-logging-and-log-levels)

dkent600:
Default is "error"

PR #137, opened 2018-03-09
added getDefaultAccountAsync

dkent600:
Addresses:  

https://github.com/daostack/arc.js/issues/134
https://github.com/daostack/arc.js/issues/16
Challenges presented by Metamask to Alchemy and Vanille.

To wit: 
- `requireContract` and `getDefaultAccount` are now async.
- using `web3.eth.getAccounts` rather than `web3.eth.accounts`
- `getDefaultAccount` now always resets web3.eth.defaultAccount
- renamed ExtendTruffleContract methods `at`, `new` and `deployed` so they won't be confused with the static truffle contract methods of the same name.
- added LoggingService


index.d.ts
```diff
@@ -4,6 +4,53 @@ import * as Web3 from "web3";
...
 declare module "@daostack/arc.js" {
+
+  /**
+   * logging levels
+   */
+  export enum LogLevel {
+    all = 15,
+    debug = 4,
+    error = 8,
+    info = 1,
```
tibetsprague: I would order these by amount of logging

dkent600: Yeah, I did it that way out of paranoia of lint.  I'll go ahead and change the order, even if I have to disable lint for that line.

Issue #118, opened 2018-03-04
Handle new Arc scheme permissions


dkent600:
See https://github.com/daostack/arc/pull/390/commits/6baf3f95f4f401e090293dc7c21627a95ffb342e

dkent600:
Also notice gas has changed: https://github.com/daostack/arc/pull/389#issuecomment-370214374

dkent600:
![](https://github.trello.services/images/mini-trello-icon.png) [Handle new Arc scheme permissions](https://trello.com/c/KwsvgyFb/102-handle-new-arc-scheme-permissions)

dkent600:
Done

Issue #38, opened 2018-01-22
Define permission values as constants


dkent600:
define "0x00000001" as a constant permission value "REGISTERED"
also define all the other permission values as constants too

tibetsprague:
![](https://github.trello.services/images/mini-trello-icon.png) [Define permission values as constants](https://trello.com/c/6ToALGyc/31-define-permission-values-as-constants)

dkent600:
Done

PR #145, opened 2018-03-14
upgrade to ArcV35

dkent600:
Addresses: #38 and #118 

- increased gas limit
- expanded permissions for SchemeRegistrar
- changed params for `SchemeRegistrar.proposeToAddOrModifyScheme` and `proposeToRemoveScheme`
- permissions are now defined in enums and passed around as such, rather than as strings
- sets gas limit in `DaoCreator.forgeOrg` when using a non-universal controller.

**Breaking**
- changed params for `SchemeRegistrar.proposeToAddOrModifyScheme` and `proposeToRemoveScheme`
- permissions are now defined in enums and passed around as such, rather than as strings (see `Dao.new`, `scheme.getDefaultPermissions`, DAO.getSchemes)

### DIFF
--- a/lib/commonTypes.ts
+++ b/lib/commonTypes.ts
@@ -1,3 +1,5 @@
+import { Utils } from "./utils";
+
 export type fnVoid = () => void;
 export type Hash = string;
 export type Address = string;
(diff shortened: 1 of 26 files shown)


lib/commonTypes.ts
```diff
@@ -33,3 +35,52 @@ export interface GetDaoProposalsConfig {
    */
   proposalId?: Hash;
 }
+
+export enum SchemePermissions {
```
tibetsprague: why again do we have to duplicate this? why cant index.d.ts import and export these?

lib/commonTypes.ts
```diff
@@ -33,3 +35,52 @@ export interface GetDaoProposalsConfig {
...
+ * be able to perform its full range of functionality.
+ *
+ * Note that '1' is always assigned to a scheme by the Controller when the
+ * scheme is registered with the controller.
+ */
+export enum DefaultSchemePermissions {
+  NoPermissions = SchemePermissions.None,
+  MinimumPermissions = SchemePermissions.IsRegistered,
+  AllPermissions = SchemePermissions.All,
+  ContributionReward = SchemePermissions.IsRegistered,
```
tibetsprague: do we really need this enum? seems like we could define the defaults in each contractwrapper class

lib/contracts/schemeregistrar.ts
```diff
@@ -88,50 +68,34 @@ export class SchemeRegistrarWrapper extends ExtendTruffleContract {
...
-
-        if (permissions > Number(this.getDefaultPermissions())) {
-          throw new Error(
-            "SchemeRegistrar cannot work with schemes having greater permissions than its own"
-          );
-        }
+      /**
+       * then we are adding/removing an Arc scheme and can get and check its permissions.
+       */
+      permissions = options.permissions || DefaultSchemePermissions[options.schemeName];
```
tibetsprague: oh I see we use it here

lib/commonTypes.ts
dkent600: We're just in an in-between stage right now.  The goal is to automatically generate index.d.ts, by whatever means.  We're just not there yet.  But getting closer...

tibetsprague: even now though you could import/export. not a big deal, just wondering why not

dkent600: 1) I only just recently finished converting the code to typescript and there has been a bunch of higher priority stuff to work on.
2) You may be right that it is easy to do, but I've found no evidence that it is as easy as you say, I've been assuming it is not.
3) So I would like to be able to concentrate on that task and not mix it in with other stuff.

Issue #148, opened 2018-03-19
use environment variables to configure migration scripts, clean up documentation


dkent600:
1. Use environment variables to configure migration scripts
2. clean up documentation on migrations ("Deploying Contracts to a Specified Network").  Make it good for both repo and application contexts.

The migrations script should use the Config service rather than reading the default.json file directly.   (This is a bit tricky since, being Typescript, the config service must be compiled first, putting making the migration dependent on a build being done.  Might need to convert Config back to JavaScript?)

dkent600:
The code part of this is addressed in https://github.com/daostack/arc.js/pull/146.

Still need to update the docs.

dkent600:
Done

PR #150, opened 2018-03-19
Refactor contracts

dkent600:
Addresses: #8, #148 (documentation)

Changes:
- renames Contracts to WrapperService
- renames Config to ConfigService
- adds the `InitializeArc` method that application must be call once before using Arc.js
- *Wrapper classes are now exposed in index.d.ts and are explicitly distinct from their factory classes.  (This has/will already happen anyway via automated type declarations)
- the new WrapperService interface is a lot nicer (IMHO) than the `contracts` class it replaces.  See the static properties `wrappers`, `wrappersByType` and `wrapperFactories`.  Note that they are synchronous.

**Note** this does nothing wrt Arc.js or contract versioning.  That is left for another time.


lib/dao.ts
```diff
@@ -48,9 +48,9 @@ export class DAO {
     return new Promise<string>(
       async (resolve: (address: Address) => void, reject: (ex: any) => void): Promise<void> => {
         try {
-          const contracts = await Contracts.getDeployedContracts();
-          const daoCreator = await DaoCreator.at(
-            daoCreatorAddress ? daoCreatorAddress : contracts.allContracts.DaoCreator.address);
+          const daoCreator =
+            daoCreatorAddress ?
+              await WrapperService.factories.DaoCreator.at(daoCreatorAddress) : WrapperService.wrappers.DaoCreator;
```
tibetsprague: this still feels pretty confusing and non intuitive to me. hmm. im fine with this refactor but i think there is a next level of clarity we can and should get to

lib/dao.ts
```diff
@@ -97,15 +97,15 @@ export class DAO {
     const controller = this.controller;
     const avatar = this.avatar;
     const arcTypesMap = new Map<Address, string>(); // <address: Address, name: string>
-    const contracts = await Contracts.getDeployedContracts();
+    const wrapperService = WrapperService.wrappersByType;
```
tibetsprague: this actually feels less clear to me. seems like outside people are going to be confused by the term "wrapper". Wrapper of what? dont they want to be interacting with the contracts?

lib/dao.ts
```diff
@@ -172,7 +172,7 @@ export class DAO {
    * @param address - optional
    */
   public async getContractWrapper(contract: string, address?: Address): Promise<ContractWrapperBase | undefined> {
-    return Contracts.getContractWrapper(contract, address);
+    return WrapperService.getContractWrapper(contract, address);
```
tibetsprague: also why call these things Service? seems a bit unnecessary of a word to me

lib/wrapperService.ts
```diff
@@ -0,0 +1,175 @@
...
+import { TokenCapGC, TokenCapGCWrapper } from "./wrappers/tokenCapGC.js";
+import { UpgradeScheme, UpgradeSchemeWrapper } from "./wrappers/upgradescheme.js";
+import { VestingScheme, VestingSchemeWrapper } from "./wrappers/vestingscheme.js";
+import { VoteInOrganizationScheme, VoteInOrganizationSchemeWrapper } from "./wrappers/voteInOrganizationScheme.js";
+
+/**
+ * An object with property names being a contract key and property value as the
+ * corresponding wrapper factory (ContractWrapperFactory<TWrapper).
+ */
+export interface ArcWrapperFactories {
```
tibetsprague: honestly this seems overly complex. maybe we should discuss a bit more in person the need for all these layers and modes of interaction with contracts

lib/wrapperService.ts
```diff
@@ -0,0 +1,175 @@
...
+  GlobalConstraintRegistrar: GlobalConstraintRegistrarWrapper;
+  SchemeRegistrar: SchemeRegistrarWrapper;
+  TokenCapGC: TokenCapGCWrapper;
+  UpgradeScheme: UpgradeSchemeWrapper;
+  VestingScheme: VestingSchemeWrapper;
+  VoteInOrganizationScheme: VoteInOrganizationSchemeWrapper;
+}
+
+/**
+ * Arc.js wrapper factores grouped by type.
```
tibetsprague: factories

lib/wrapperService.ts
```diff
@@ -0,0 +1,175 @@
...
+   * It is currently called in ArcInitialize(), which in trun must be invoked by any applicaiton
+   * using Arc.js.
+   */
+  public static async initialize(): Promise<void> {
+    LoggingService.debug("WrapperService: initializing");
+    /**
+     * Deployed contract wrappers by name.
+     */
+    WrapperService.wrappers = {
+      AbsoluteVote: await AbsoluteVote.deployed(),
```
tibetsprague: why do we need to load all the deployed contracts at the beginning? seems unnecessary since most apps wont use all of them.

lib/wrapperService.ts
```diff
@@ -0,0 +1,175 @@
...
+      ],
+      votingMachines: [
+        WrapperService.wrappers.AbsoluteVote,
+        WrapperService.wrappers.GenesisProtocol,
+      ],
+    };
+  }
+
+  /**
+   * Returns an Arc.js contract wrapper or undefined if not found.
```
tibetsprague: actually it returns a promise no?

docs/Configuration.md
```diff
@@ -34,15 +34,15 @@ The level of logging.  Default is 8 (error).  The available log levels, which ma
 ### Obtain a configuration setting at runtime
 
 ```javascript
-import { Config } from '@daostack/arc.js';
-Config.get('network');
+import { ConfigService } from '@daostack/arc.js';
```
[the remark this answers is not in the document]
dkent600: I'm trying to follow a consistent naming convention for classes that are effectively services.  They probably ought to all be in a Services folder.

lib/dao.ts
dkent600: I am following a common pattern to define and name these classes according to their functionality as a "Service".

lib/wrapperService.ts
dkent600: I thought a lot about that.  The alternative is to load on demand.  That would mean that the WrapperService would need to be instantiated, and then something like 

```
const upgradeScheme = WrapperService.wrapper.UpgradeScheme;
```

would become:  

```
const wrapperService  = new WrapperService();
const upgradeScheme = await wrapperService.getScheme("UpgradeScheme");
```

I chose to go with the cleaner usage.

Another option could be that in `ArcInitialize()` we add a parameter that is a collection of contract names that the application intends to use, and filter what we load by that list.  Easy to do.  Similar to other libraries that allow one to customize the footprint of the package.

lib/wrapperService.ts
dkent600: yeah, I tend not too mention that it is a promise in those comments.  maybe I should.

lib/wrapperService.ts
dkent600: The typings work nicely, and the resulting code, as used by the consumer, is very clean.

lib/dao.ts
dkent600: The docs describe wrappers right up front as wrapping Arc contracts.

dkent600: An alternative I've been seriously considering:

Rename:

UpgradeScheme =>  UpgradeSchemeFactory
UpgradeSchemeWrapper => UpgradeScheme

docs/Home.md
```diff
@@ -279,136 +294,123 @@ The DAOstack DAO is named "Genesis".  Obtain it like this:
...
-#### Voting Machines
-* AbsoluteVote
-* GenesisProtocol
+### How to obtain a contract wrapper
+Arc.js provides several ways to obtain a contract wrapper:
 
-#### Global Constraints
+1. Using the `WrapperService.wrappers` to obtain the wrappers of the contracts as deployed by the current version of Arc.js.
+2. Using the wrapper's factory class and calling `at(someAddress)`, `new()` and `deployed()`.
+3. Using `WrapperService.getContractWrapper` or  `DAO.getContractWrapper`
```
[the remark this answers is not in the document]
tibetsprague: different use cases would be great

lib/dao.ts
dkent600: Something like this might be good:

```
const daoCreator = daoCreatorAddress ? 
   await ContractWrappers.DaoCreator.factory.at(daoCreatorAddress) : 
   ContractWrappers.DaoCreator; 
```

docs/Home.md
```diff
@@ -279,136 +294,123 @@ The DAOstack DAO is named "Genesis".  Obtain it like this:
...
 
-#### Voting Machines
-* AbsoluteVote
-* GenesisProtocol
+### How to obtain a contract wrapper
+Arc.js provides several ways to obtain a contract wrapper:
 
-#### Global Constraints
+1. Using the `WrapperService.wrappers` to obtain the wrappers of the contracts as deployed by the current version of Arc.js.
+2. Using the wrapper's factory class and calling `at(someAddress)`, `new()` and `deployed()`.
```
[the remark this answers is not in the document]
dkent600: They are explained in the sections immediately following.

docs/Home.md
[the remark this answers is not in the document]
dkent600: They are explained in the sections immediately following.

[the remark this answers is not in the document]
dkent600: Our comments crossed on the  wire.  How about making sure we've settled on the architecture here first (see my suggestions elsewhere for possible changes)

Issue #96, opened 2018-02-19
automatically generate d.ts (typings)


dkent600:
Once we have all the code in typescript (#58), and the code is fully fleshed out with data types and return type definitions (#95), we should automate the generation of the d.ts file.  This will reduce the potential for errors arising from redundancy.

dkent600:
![](https://github.trello.services/images/mini-trello-icon.png) [automate generation of d.ts (typings)](https://trello.com/c/YxBwjYVl/87-automate-generation-of-dts-typings)

dkent600:
I've looked at several packages trying to find a tool that will generate a d.ts in the structure that we use.  I've posed a question on stackoverflow.  All to no avail, as yet.

What I do have is a set of steps to do this manually.  The process takes only a couple minutes.  I could probably write some code to automated this.

The steps:
```
1. Run:

	type dist/*.d.ts > index.d.ts
	type dist/contracts/*.d.ts >> index.d.ts

2. Remove the arc.ts imports at top.

3. Change all: "declare " to: ""

4. Remove all statements that start with "import "

5. Surround everything with:

	import * as BigNumber from "bignumber.js";
	declare module "@daostack/arc.js" {
	
		[everything]
	
	}
```

tibetsprague:
yikes, thats crazy. what are other people doing?

dkent600:
dunno, that is why I posed the question on stackoverflow.

dkent600:
Done.  turned out to be easy (see the stackoverflow question)

PR #152, opened 2018-03-21
Autogenerate typings

dkent600:
Addresses: https://github.com/daostack/arc.js/issues/96

- renames arc.ts to index.ts
- generates d.ts for all modules, including the new index.ts, at build time
- associated package.json changes
- some typings corrections

**Breaking**
The distinction between wrapper factories and wrapper classes is now exported.  Implication, for example:

`import {UpgradeScheme} from "@daostack/arc.js"` now only pulls in the factory class.  The manually-generated typings were able to fake that this was not also the wrapper class by merging the definition of factory and wrapper classes into a single "UpgradeScheme" typedef.  This merging is no longer happening.

So now to use the wrapper you have to do:

`import {UpgradeSchemeWrapper} from "@daostack/arc.js"`


tsconfig.json
```diff
@@ -6,7 +6,8 @@
     "preserveConstEnums": true,
     "noUnusedLocals": true,
     "sourceMap": true,
-    "baseUrl": "lib"
+    "baseUrl": "lib",
+    "declaration": true
```
tibetsprague: this was all you needed? nice :)

Issue #156, opened 2018-03-23
migrateContracts doesn't work in app context without npm install


dkent600:
The migrateContracts script doesn't work in the context of an application without doing a full `npm install` in the folder of the installed Arc.js package.  

What happens:

> Error: Options {"loose":true} passed to C:\Users\dkent\Documents\GitHub\Vanille\node_modules\babel-preset-env\lib\index.js which does not accept options. (While processing preset: "C:\\Users\\dkent\\Documents\\GitHub\\Vanille\\node_modules\\babel-preset-env\\lib\\index.js") (While processing preset: "C:\\Users\\dkent\\Documents\\GitHub\\Vanille\\node_modules\\babel-preset-env\\lib\\index.js")

But that is a worthless message.  What is really happening is that while `require` is being passed the correct path to `babel-plugin-env`, it nevertheless returns an empty object (`{}`), completely ignoring the code that is present at that file location.

This was happening for me in two projects and on two machines (my native and my VM)?

`require` will return an empty object when there is a circular dependency.  But I couldn't find one, though I wouldn't bet much on the quality of the tools I used.

Next step might be to figure out what happened between release 39 and 40.  39 did not have this problem.

In the meantime I am working around this problem by enforcing a full `npm install` in the folder of the installed Arc.js package the first time that `migrateContracts` is run, which is a bit annoying, but works.

dkent600:
One thing worth investigating is that `truffle-core-migrate-without-compile` pulls in a deprecated preset for doing transformations.  Just maybe that is related.

(Hmmm, seems like it doesn't invoke babel at all unless it is running as a package........)

dkent600:
![](https://github.trello.services/images/mini-trello-icon.png) [Fix issue with migrateContracts scripts in context of applications](https://trello.com/c/FEjPvXeO/108-fix-issue-with-migratecontracts-scripts-in-context-of-applications)

dkent600:
no longer relevant

PR #165, opened 2018-03-26
Changes to GenesisProposal parameters

### LINKED ISSUE #159: Is threshold const A too small in migration?
In the migration we set `threshold cost A` to be 2, this means that if there are no boosted proposals, the threshold is exactly 2 **in wei** (meaning the smallest unit of the `DAOToken`). 
This seems very low and maybe a mistake.

In alchemy for example, we stake in units of `GEN`, which are 10^18 wei. So the threshold is practically almost zero.

### LINKED ISSUE #162: change types of thresholdCostA ,proposingRepRewardConstB & proposingRepRewardConstA
GenesisProtocolParams.thresholdCostA, proposingRepRewardConstB and proposingRepRewardConstA should be defined as bignumbers|string, not number.

See https://github.com/daostack/arc.js/issues/159#issuecomment-376107675

### LINKED ISSUE #163: Export ContributionRewardWrapper.RewardType
So it can be used in apps

dkent600:
Resolves #159 
Resolves #163 
Resolves #162

Refer to: https://github.com/daostack/arc.js/issues/159#issuecomment-376107675

### DIFF
--- a/lib/wrappers/genesisProtocol.ts
+++ b/lib/wrappers/genesisProtocol.ts
@@ -869,7 +869,8 @@ export class GenesisProtocolWrapper extends ContractWrapperBase implements Schem
       },
       params);
 
-    const maxEthValue = 10 ** 26;
+    const maxEthValue = web3.toBigNumber(10).pow(26);
+
     const proposingRepRewardConstA = web3.toBigNumber(params.proposingRepRewardConstA);
 
     if (proposingRepRewardConstA.lt(0)) {
--- a/lib/wrappers/genesisProtocol.ts
+++ b/lib/wrappers/genesisProtocol.ts
@@ -996,17 +996,17 @@ export interface GenesisProtocolParams {
    */
   preBoostedVotePeriodLimit: number;
   /**
-   * the time limit in seconds for a proposal to be in an relative voting mode.
+   * The time limit in seconds for a proposal to be in an relative voting mode.
    * Default is 604800 (one week).
    */
   boostedVotePeriodLimit: number;
   /**
-   * Constant A in the threshold calculation. See [[GenesisProtocol.getThreshold]].
+   * Constant A in the threshold calculation. See [[GenesisProtocolWrapper.getThreshold]].
    * Default is 2, converted to Wei
    */
   thresholdConstA: BigNumber.BigNumber | string;
   /**
-   * Constant B in the threshold calculation. See [[GenesisProtocol.getThreshold]].
+   * Constant B in the threshold calculation. See [[GenesisProtocolWrapper.getThreshold]].
    * Default is 10
    */
   thresholdConstB: number;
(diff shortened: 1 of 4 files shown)


lib/wrappers/genesisProtocol.ts
```diff
@@ -852,43 +854,65 @@ export class GenesisProtocolWrapper extends ContractWrapperBase implements Schem
 
     params = Object.assign({},
       {
-        boostedVotePeriodLimit: 60,
+        boostedVotePeriodLimit: 604800, // 1 week
```
orenyodfat: These are default values which can be override be arc.js user ?

orenyodfat: Does these default values are the same values used by the migration script ?
If yes, so I suggest to define that externally (e.g GenesisProtocol default params or something) and import/use both here and in the migration script

lib/wrappers/genesisProtocol.ts
```diff
@@ -961,138 +985,70 @@ export interface RedeemEventResult {
 
 export interface GenesisProtocolParams {
   /**
-   * the absolute vote percentages bar
-   * Must be greater than zero.
+   * The percentage of the absolute vote that must be exceeded to result in a win.
+   * Must be between 0 and 100.
    * Default is 50.
```
orenyodfat: Again , define the defaults values  externally  and point from the comment to the defaults definition..
By doing that you , there will be no need to update all comments when updating the defaults.

orenyodfat:
Travis is red

lib/wrappers/genesisProtocol.ts
```diff
@@ -852,43 +854,65 @@ export class GenesisProtocolWrapper extends ContractWrapperBase implements Schem
...
     }
 
-    if ((params.preBoostedVoteRequiredPercentage <= 0) || (params.preBoostedVoteRequiredPercentage > 100)) {
+    const preBoostedVoteRequiredPercentage = web3.toBigNumber(params.preBoostedVoteRequiredPercentage);
+
+    if (preBoostedVoteRequiredPercentage.lte(0) || preBoostedVoteRequiredPercentage.gt(100)) {
       throw new Error("preBoostedVoteRequiredPercentage must be greater than 0 and less than or equal to 100");
     }
 
+    const stakerFeeRatioForVoters = web3.toBigNumber(params.stakerFeeRatioForVoters);
```
tibetsprague: hmm, why converting the ratios to a big number? that seems wrong

lib/wrappers/genesisProtocol.ts
dkent600: Yes

dkent600:
@orenyodfat @tibetsprague 

> Travis is red

That is because I don't yet have the corresponding changes from Arc.  I wasn't actually intending for the PR to be reviewed yet (hadn't invited anyone).  I'm not quite sure what the process is for accomplishing that.

lib/wrappers/genesisProtocol.ts
dkent600: The comments are already only in one place.  I don't worry about the comments in the migration script.

But still, your suggestion is one that I have also considered several times.  It depends on the defaults for DAO.new being necessarily the same as the defaults for the Genesis DAO.  For now at least, that is the case, but may not always be.

In any case, there are issues with including code in the deployment script, and babel: (https://github.com/daostack/arc.js/issues/156).  I would have to figure out how to avoid those issues, and it is a pain to do as it requires testing migration in the context of an application.

There could also result problems with auto-generated API documentation, depending on how I solve the problem with babel.

So I am not inclined right now to make this change.

lib/wrappers/genesisProtocol.ts
dkent600: So response below

lib/wrappers/genesisProtocol.ts
orenyodfat: is it no as simple as creating a file (naming something like genesisDefaultParams.js and import that fro the migration file ?

dkent600:
Closing until we have the required update from Arc.

lib/wrappers/genesisProtocol.ts
dkent600: No, it isn't that simple, it has the implications I described above.  However, I think it is worth figuring out how to do this.  There is another bit that has also become redundant because of this problem, and I don't like it either.  So I've created a ticket on it:  https://github.com/daostack/arc.js/issues/171

dkent600: One option that would be easy would be to put the data into a .json file rather than javascript.

Issue #171, opened 2018-03-28
Enabled importing code into the migration script


dkent600:
Enable importing code into the migration script, to avoid redundant data.

See these comments:  https://github.com/daostack/arc.js/pull/165#discussion_r177429498

dkent600:
One option that would be easy would be to put the data into a .json file rather than javascript.

dkent600:
Finally figured out how to work-around this in a way that allows code to be imported into the migration script.  The work-around is to remove the following code from truffle.js:

```
require("babel-polyfill");
require("babel-register")({
  "presets": ["es2015"],
  "plugins": ["syntax-async-functions", "transform-regenerator"]
});
```

I don't know what that was there.  Perhaps this limits the migration scripts to ES5.

Issue #128, opened 2018-03-07
Increase gasLimit for forgeOrg


dkent600:
Per @orenyodfat 

Please note that forgeOrg consume more gas as the number of founders increase.
My suggestion is to set 'gasLimit-deployment' for this function to be on the safe side.

dkent600:
Note this change could result in hassles with Metamask....

dkent600:
@orenyodfat With respect to gas, how many founders should we be able to handle?

orenyodfat:
you should try and see :) .

dkent600:
@orenyodfat  Just did forty founders, no problem.  Ganache, 1000 tokens, 1000 reputation

dkent600:
@orenyodfat I am inclined to think that when creating a DAO, in order to support the maximum possible number of founders, we should dynamically determine the current network block gas limit and supply that.  (I believe this would be easy to do.)

What do you think?

The only remaining question, I think, would then be to communicate to users, via documentation or at runtime in the app, how many founders they can expect to be able to create, given the current block gas limit.

orenyodfat:
The code should look like 
> var gasPerFounder = 50000;
var gas = founders.length *gasPerFounder + gasLimit;
if gas > mainNetGasLimt {
  gas = mainNetGasLimt;
}
var tx = await daoCreator.forgeOrg("testOrg","TEST","TST",founders,foundersToken,foundersReputation,0,{gas: gas});

dkent600:
![](https://github.trello.services/images/mini-trello-icon.png) [Increase gasLimit for forgeOrg](https://trello.com/c/gWpuGRb8/114-increase-gaslimit-for-forgeorg)

dkent600:
![](https://github.trello.services/images/mini-trello-icon.png) [Increase gasLimit for forgeOrg](https://trello.com/c/gWpuGRb8/114-increase-gaslimit-for-forgeorg)

PR #172, opened 2018-03-29
Changes to GenesisProtocol parameters

### LINKED ISSUE #159: Is threshold const A too small in migration?

### LINKED ISSUE #162: change types of thresholdCostA ,proposingRepRewardConstB & proposingRepRewardConstA

### LINKED ISSUE #163: Export ContributionRewardWrapper.RewardType

dkent600:
Resolves #159
Resolves #163
Resolves #162

- Upgrades to Arc v38.
- switches Genesis DAO to the universal controller

Refer to: 

- #159 (comment)
- https://github.com/daostack/arc.js/pull/165 (previous closed PR on this)

**Breaking**: Converts number-typed values returned by `getParameters` and `getThreshold` from `BigNumber` to `number`

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -1,7 +1,7 @@
 {
   "autoApproveTokenTransfers": true,
   "defaultVotingMachine": "AbsoluteVote",
-  "gasLimit_deployment": 5500000,
+  "gasLimit_deployment": 5700000,
   "gasLimit_runtime": 4543760,
   "network": "ganache",
   "providerUrl": "http://127.0.0.1",
--- a/lib/wrappers/daocreator.ts
+++ b/lib/wrappers/daocreator.ts
@@ -300,7 +300,7 @@ export interface ForgeOrgConfig {
    */
   name: string;
   /**
-   * A cap on the number of tokens, in DAO's token.  Default is zero, which means no cap.
+   * A cap on the number of tokens, in the DAO's token.  Default is zero, which means no cap.
    */
   tokenCap: BigNumber.BigNumber | string;
   /**
--- a/lib/wrappers/daocreator.ts
+++ b/lib/wrappers/daocreator.ts
@@ -300,7 +300,7 @@ export interface ForgeOrgConfig {
    */
   name: string;
   /**
-   * A cap on the number of tokens.  Default is zero.
+   * A cap on the number of tokens, in DAO's token.  Default is zero, which means no cap.
    */
   tokenCap: BigNumber.BigNumber | string;
   /**
--- a/lib/wrappers/daocreator.ts
+++ b/lib/wrappers/daocreator.ts
@@ -39,6 +39,7 @@ export class DaoCreatorWrapper extends ContractWrapperBase {
     const defaults = {
       founders: [],
       name: undefined,
+      tokenCap: 0,
       tokenName: undefined,
       tokenSymbol: undefined,
       universalController: true,
@@ -76,6 +77,7 @@ export class DaoCreatorWrapper extends ContractWrapperBase {
       options.founders.map((founder: FounderConfig) => web3.toBigNumber(founder.tokens)),
       options.founders.map((founder: FounderConfig) => web3.toBigNumber(founder.reputation)),
       controllerAddress,
+      options.tokenCap,
       /**
        * We need to increase the gas limit when creating for a non-universal controller,
        * or it will revert.  MetaMask will probably complain that our gas exceeds the block limit,
@@ -297,6 +299,10 @@ export interface ForgeOrgConfig {
    * The name of the new DAO.
    */
   name: string;
+  /**
+   * A cap on the number of tokens.  Default is zero.
+   */
+  tokenCap: BigNumber.BigNumber | string;
   /**
    * The name of the token to be associated with the DAO
    */
--- a/migrations/2_deploy_organization.js
+++ b/migrations/2_deploy_organization.js
@@ -78,7 +78,7 @@ module.exports = async (deployer) => {
       founders.map((f) => web3.toWei(f.reputation)),
       // use non-universal controller
       0,
-      100000000, // token cap of one hundred million GEN
+      web3.toWei(100000000), // token cap of one hundred million GEN, in Wei
       { gas: gasAmount });
 
     const AvatarInst = await Avatar.at(tx.logs[0].args._avatar);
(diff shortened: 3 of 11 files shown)


lib/wrappers/daocreator.ts
```diff
@@ -297,6 +299,10 @@ export interface ForgeOrgConfig {
    * The name of the new DAO.
    */
   name: string;
+  /**
+   * A cap on the number of tokens.  Default is zero.
```
orenyodfat: will comment also that zero = no cap.

test/genesisProtocol.js
```diff
@@ -151,8 +151,8 @@ describe("GenesisProtocol", () => {
       avatar: dao.avatar.address,
     });
     assert.isOk(result);
-    assert.equal(result.thresholdConstA, 1);
-    assert.equal(result.thresholdConstB, 1);
+    assert.equal(web3.fromWei(result.thresholdConstA).toNumber(), 2);
```
orenyodfat: from where 2 and 10 come from ?
Would it will be easier to import this default values from somewhere..as when someday it will be changed this code can be relax?

migrations/2_deploy_organization.js
```diff
@@ -42,12 +42,12 @@ module.exports = async (deployer) => {
     preBoostedVoteRequiredPercentage: 50,
     preBoostedVotePeriodLimit: 5184000, // 2 months
     boostedVotePeriodLimit: 604800, // 1 week
-    thresholdConstA: 2,
+    thresholdConstA: web3.toWei(2),
```
orenyodfat: I suggest to define GP default values separately and import it here,

orenyodfat:
The following log (from test) does not looks good.

> Arc.js (error): hydrateFromAt failing: Error: Cannot create instance of UpgradeScheme; no code at address 0x0000000000000000000000000000000000000000

config/default.json
```diff
@@ -1,7 +1,7 @@
 {
   "autoApproveTokenTransfers": true,
   "defaultVotingMachine": "AbsoluteVote",
-  "gasLimit_deployment": 5500000,
+  "gasLimit_deployment": 5700000,
```
orenyodfat: How do you get this gas number ?
How about fetching the gas limit from arc package?

test/genesisProtocol.js
dkent600: Yes, see https://github.com/daostack/arc.js/issues/171

Will probably create a json file for these values.

migrations/2_deploy_organization.js
dkent600: See https://github.com/daostack/arc.js/issues/171

config/default.json
dkent600: The Arc value isn't enough for founders.  This is enough for our currently nine founders on Kovan.

See https://github.com/daostack/arc.js/issues/128

dkent600:
@orenyodfat 

> The following log (from test) does not looks good.
>> Arc.js (error): hydrateFromAt failing: Error: Cannot create instance of UpgradeScheme; no code at address 0x0000000000000000000000000000000000000000

It is a normal outcome of the test which is exercising what happens when we try to create a wrapper on a contract at a null address.  It is an error-level logging message.

config/default.json
dkent600: (I've got this card as the top priority, under a documentation card)

test/genesisProtocol.js
orenyodfat: why not including that on this pr ?

migrations/2_deploy_organization.js
orenyodfat: why not including that on this pr ?

config/default.json
orenyodfat: You should take the gas limit from arc (which is enough for 3 founders) and then implement 128.

test/genesisProtocol.js
dkent600: >why not including that on this pr ?

Trying to keep tickets simpler, isolate the work, and so I can focus on the task.  Ideally I would be able to import code, not just json, but that is going to take focus and time.

orenyodfat:
>It is a normal outcome of the test which is exercising what happens when we try to create a wrapper on a contract at a null address. It is an error-level logging message.

Why this test is testing ? why it is needed ?

migrations/2_deploy_organization.js
dkent600: > why not including that on this pr ?

See response above.

dkent600:
>> It is a normal outcome of the test which is exercising what happens when we try to create a wrapper on a contract at a null address. It is an error-level logging message.

> Why this test is testing ? why it is needed ?

To confirm that the method returns `undefined`, as it is spec'd to do (see the test).

test/genesisProtocol.js
orenyodfat: it suppose to be part of this pr.
as this pr change GP params. why double work ?

orenyodfat:
>To confirm that the method returns undefined on invalid addresses, as it is spec'd to do.

can you just check this address validity before using that to avoid this unpleasant error log ?

dkent600:
> To confirm that the method returns undefined on invalid addresses, as it is spec'd to do.
can you just check this address validity before using that to avoid this unpleasant error log ?

What I did was to turn off logging just for that test.

test/genesisProtocol.js
dkent600: the "double work" is trivial.  #171 explains why the alternative is not so trivial (particularly when importing code).

config/default.json
dkent600: Good to know that the Arc limit accounts for 3 founders.  I suggest that the work to do the runtime computation for `forgeOrg` (#128) be done in a separate PR.

PR #181, opened 2018-04-04
dynamic gas computation and importing code in migration scripts

dkent600:
Resolves: #128 #171 #182 

- Dynamically compute amount of gas needed to create/migrate a DAO
- Enable us to import code in the migration script, which is now written in TypeScript
- consolidates duplicate definitions of data
- all gas limits are stored in a new `gasLimits.js` file, no longer among the config settings
- deployment script is now a TypeScript file, stored under lib/migrations and built along with everything else.
- typings error in `dao.new`


dkent600:
Not ready

docs/Migration.md
```diff
@@ -3,9 +3,9 @@
 To deploy contracts to a specified network, follow these steps:
 
 1. Set the environment variable `arcjs_network` to the name of the network ("ganache", "kovan", "ropsten", "live"). The default is "ganache".  Truffle will use this setting to find the specified settings in truffle.js in the Arc.js package.  The migration script will also use this setting just to feedback to you which network truffle is using.
-2. The migration script will use `arcjs_gasLimit_deployment`, allowing you to customize the gas amount if needed.
-3. make sure that /migrations/founders.json in the Arc.js package folder has an entry for your test network with the founders to add to the Genesis DAO.
-4. make sure you have your testnet running and listening on the right port.
+2. The migration script will use the gas settings defined the Arc.js file `/gasLimits.js`.  The gas limit when migrating/creating Daos is computed dynamically as a function of the number of founders.
+3. Make sure that `/migrations/founders.json` in the Arc.js package folder has an entry for your test network with the founders to add to the Genesis DAO.
```
tibetsprague: wait, so to deploy to ganache now we have to add a founders.json file to Arc.js? that seems really weird

docs/Migration.md
```diff
@@ -3,9 +3,9 @@
 To deploy contracts to a specified network, follow these steps:
 
 1. Set the environment variable `arcjs_network` to the name of the network ("ganache", "kovan", "ropsten", "live"). The default is "ganache".  Truffle will use this setting to find the specified settings in truffle.js in the Arc.js package.  The migration script will also use this setting just to feedback to you which network truffle is using.
-2. The migration script will use `arcjs_gasLimit_deployment`, allowing you to customize the gas amount if needed.
-3. make sure that /migrations/founders.json in the Arc.js package folder has an entry for your test network with the founders to add to the Genesis DAO.
-4. make sure you have your testnet running and listening on the right port.
+2. The migration script will use the gas settings defined the Arc.js file `/gasLimits.js`.  The gas limit when migrating/creating Daos is computed dynamically as a function of the number of founders.
```
tibetsprague: settings define *in* the

config/default.json
```diff
@@ -1,8 +1,7 @@
 {
   "autoApproveTokenTransfers": true,
   "defaultVotingMachine": "AbsoluteVote",
-  "gasLimit_deployment": 5700000,
-  "gasLimit_runtime": 4543760,
+  "foundersInGasLimitArc": 3,
```
tibetsprague: why does this need to be configurable if you cant configure the gas limits themselves?

docs/Migration.md
dkent600: No, you don't have to add it, it is there just as before, nothing has changed in that respect.  I will add a comment that founders for ganache are already in it.

config/default.json
dkent600: 'foundersInGasLimitArc' should not still have been in there.  I have removed it.

This config file no longer relates to gas limits at all, `gasLimit.js` has replaced it in that respect, and technically one can modify that file to change the gas limits, though it is not as handy as environment variables.  I don't see much need for customizing gas limits so I think removing the ability won't be missed.

docs/Migration.md
```diff
@@ -3,8 +3,8 @@
...
-2. The migration script will use the gas settings defined the Arc.js file `/gasLimits.js`.  The gas limit when migrating/creating Daos is computed dynamically as a function of the number of founders.
-3. Make sure that `/migrations/founders.json` in the Arc.js package folder has an entry for your test network with the founders to add to the Genesis DAO.
+2. The migration script will use the gas settings defined in the Arc.js file `/gasLimits.js`.  The gas limit when migrating/creating Daos is computed dynamically as a function of the number of founders.
+3. Make sure that `/migrations/founders.json` in the Arc.js package folder has an entry for your test network with the founders to add to the Genesis DAO.  Founders for ganache and kovan are already contained, though note that the kovan founders are used by DAOstack -- you might want to change those.
```
tibetsprague: so to change the founders you have to go into the node_modules folder and change this file? I still think you should be able to create and pass in your own founders file

Issue #187, opened 2018-04-06
Latest arc.js breaks external injection of web3


tibetsprague:
This prevents both interaction with MetaMask and setting up an external web3 with a different provider (like Infura) from working. The line that breaks it is in utils.ts:

`let web3: any;`

PR #188, opened 2018-04-06
fix access to globally-defined web3

dkent600:
Resolves: #187

### DIFF
--- a/custom_typings/system.d.ts
+++ b/custom_typings/system.d.ts
@@ -1,6 +1,5 @@
 declare module "system" {
   global {
-    interface Window { web3: any; }
     var window: Window;
     var artifacts: any;
   }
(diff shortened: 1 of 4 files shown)


lib/utils.ts
```diff
@@ -57,13 +53,23 @@ export class Utils {
 
     let preWeb3;
 
-    if (typeof web3 !== "undefined") {
+    let globalWeb3;
+
+    if ((typeof global !== "undefined") && (global as any).web3) {
```
tibetsprague: Why not just define web3 in the global namespace as you used to?

custom_typings/system.d.ts
```diff
@@ -1,6 +1,5 @@
 declare module "system" {
   global {
-    interface Window { web3: any; }
```
tibetsprague: Why remove this?

lib/utils.ts
dkent600: I want to undefine `web3` as global because so the TypeScript compiler will enforce the use of Utils.getWeb3().  `web` is not always accessible globally.

custom_typings/system.d.ts
dkent600: See comment above

Issue #189, opened 2018-04-06
enable applications to provide custom set of founders to migration script


dkent600:
Enable applications to provide custom set of founders to migration script.

Currently the only mechanism is to go into node_modules and modify the founders.json file in Arc.js.

dkent600:
![](https://github.trello.services/images/mini-trello-icon.png) [enable applications to provide custom set of founders to migration script](https://trello.com/c/CVtHSTaw/119-enable-applications-to-provide-custom-set-of-founders-to-migration-script)

PR #190, opened 2018-04-06
added foundersConfigurationLocation config setting

dkent600:
Resolves: #189

- Founders for networks other-than-ganache must be supplied via the `foundersConfigurationLocation` configuration setting.
- the Arc.js kovan founders are no longer stored in git
- configuration settings given in the OS environment must be pre-fixed by "arc_js".

**Question**
Should token and reputation amounts be given in Wei?

**Breaking**

Configuration settings given in the OS environment *must* be pre-fixed by "arc_js".  The code was allowing this to be optional, but we don't want to risk erroneously taking env variables from other contexts

### DIFF
--- a/lib/migrations/2_deploy_organization.ts
+++ b/lib/migrations/2_deploy_organization.ts
@@ -1,12 +1,31 @@
+import Web3 = require("web3");
+import { computeGasLimit } from "../../gasLimits.js";
 import { DefaultSchemePermissions, SchemePermissions } from "../commonTypes";
 import { ConfigService } from "../configService";
 import { GetDefaultGenesisProtocolParameters } from "../wrappers/genesisProtocol";
-const computeGasLimit = require("../../gasLimits.js").computeGasLimit;
 
+/* tslint:disable:no-console */
+/* tslint:disable:max-line-length */
+
+interface FounderSpec {
+  /**
+   * Founders' address
+   */
+  address: string;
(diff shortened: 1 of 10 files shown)


lib/migrations/2_deploy_organization.ts
```diff
@@ -1,21 +1,55 @@
...
   const network = ConfigService.get("network") || "ganache";
-  const founders = require("../../migrations/founders.json").founders[network];
+
+  const internalFoundersConfigLocation = "../../migrations/founders.json";
+  const foundersConfig = require(internalFoundersConfigLocation).founders;
+
+  const customFoundersConfigLocation = ConfigService.get("foundersConfigurationLocation");
+
+  if (internalFoundersConfigLocation !== customFoundersConfigLocation) {
+    console.log(`merging custom founders from ${customFoundersConfigLocation}`);
```
tibetsprague: seems like you should overwrite, not merge

dkent600: this does overwrite

tibetsprague: that would merge the founders from the foundersConfig with the customFoundersConfig. If you pass in a custom founders config why would you also want the default/internal founders?

tibetsprague: ok

Issue #161, opened 2018-03-25
Add parameter to InitializeArc to filter schemes


dkent600:
Add a parameter to `InitializeArc` that would filter the which contracts are loaded by `WrapperService`.

The parameter would look like this:

```
{
 include: [
    {
      name: string,
      at?: Address
    }
  ]
}
```

BTW, also rename `InitializeArc` to `InitializeArcJs`.

dkent600:
![](https://github.trello.services/images/mini-trello-icon.png) [Add parameter to InitializeArc to filter schemes](https://trello.com/c/AugELYeJ/110-add-parameter-to-initializearc-to-filter-schemes)

PR #179, opened 2018-04-03
Documentation

dkent600:
Resolves: #149 

Here is a next big iteration in Arc.js documentation.  It fills in some big gaps that were sitting out there.  There remains a big gap which is more about Proposals.  But the code relative to Proposals is also not yet filled-out nor coherent in the code, so it is not at this point feasible to write much about it.

You can see these docs here:  https://daostack.github.io/arc.js/Home/

PR #191, opened 2018-04-07
Improvements to InitializeArc

dkent600:
Resolves: #161 

- InitializeArc is renamed to InitializeArcJs
- InitializeArcJs detects when web3 is not connected
- InitializeArcJs now traps exceptions and throws a better message
- InitializeArcJs now accepts a filter of contract names to load
- LoggingService now accepts multiple loggers
- new `LoggingService.message` function for logging to multiple log levels
- change to tests so they run faster (don't call InitializeArcJs on every test)

**Note**: Documentation changes will be done in the [pending documentation PR](https://github.com/daostack/arc.js/pull/179).

**Breaking**
InitializeArc is renamed to InitializeArcJs

### DIFF
--- a/lib/wrapperService.ts
+++ b/lib/wrapperService.ts
@@ -145,8 +145,8 @@ export class WrapperService {
      * Deployed contract wrappers by name.
      */
     const filter = (options && options.filter) ?
-      Object.assign({}, WrapperService.completeWrapperFilter, options.filter) :
-      WrapperService.defaultWrapperFilter;
+      Object.assign({}, WrapperService.noWrappersFilter, options.filter) :
+      WrapperService.allWrappersFilter;
 
     /* tslint:disable:max-line-length */
     WrapperService.wrappers.AbsoluteVote = filter.AbsoluteVote ? await AbsoluteVoteFactory.deployed() : null;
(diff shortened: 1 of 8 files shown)


lib/wrapperService.ts
```diff
@@ -227,6 +238,57 @@ export class WrapperService {
...
+    GenesisProtocol: true,
+    GlobalConstraintRegistrar: true,
+    SchemeRegistrar: true,
+    TokenCapGC: true,
+    UpgradeScheme: true,
+    VestingScheme: true,
+    VoteInOrganizationScheme: true,
+  };
+
+  private static completeWrapperFilter: WrapperFilter = {
```
tibetsprague: complete is a slightly confusing word here. something that indicates  no wrappers, hmm...

dkent600:
@tibetsprague 

> Im not sure you need to have the ability to pass in multiple loggers,

I think you are talking about the new `message` function, not the actual fact that one can have multiple loggers.  Two different things.

I thought several times about the hierarchical level design you described, but everytime I come to the conclusion that I don't like it.  The hierarchical design is inflexible, and I think the importance of each level implicit in the hierarchy is artificial.  The way I've done it gives the user much more control, and it is quite clear how it works..

A future enhancement of this might be enable the log level of each logger be configured separately.

lib/wrapperService.ts
dkent600: I renamed them.

Issue #183, opened 2018-04-05
Alchemy needs someway to tell user how many transactions are left to sign for an action


tibetsprague:
This means Arc.js having some way to tell the caller how many transactions will be created for a function call, and to have callback that can be passed into every function that is called after each transaction completes to update on progress. Thoughts on the best way to implement?

dev-matan-tsuberi:
@tibetsprague Do you see this being useful for things other than staking?

tibetsprague:
anything with multiple transactions, for example I'm moving to use Arc.js's DAO creation which has 4 transactions in one. so yes, over time more and more

dkent600:
> Thoughts on the best way to implement?

This can be a new property on the object that is returned by every Arc.js that generates a transaction.

As for computing the actual count returned, I don't know if there is a way to do this dynamically, I suspect there is not.  @orenyodfat @leviadam 

So we may have to hardcode the count returned.

dkent600:
![](https://github.trello.services/images/mini-trello-icon.png) [Alchemy needs some way to track transactions](https://trello.com/c/uJ3cYvyq/121-alchemy-needs-some-way-to-track-transactions)

dkent600:
OK, tentative plan after discussion with @tibetsprague :

1. export a static object that contains the number of transactions per Arc.js operation
2. export a pub/sub interface for tracking transaction completion per operation.

dkent600:
Step 1. of the above plan was scrapped, replaced by an initial "kick-off" event that reports the expected number of transactions.

PR #195, opened 2018-04-12
Track transactions

dkent600:
Resolves: #183 

- Adds and uses `EventService` and `TransactionService` to enable tracking transactions as they complete
- removes use of and dependency on `dopts`


dev-matan-tsuberi:
1. Can I subscribe to just one `invocationKey` that i'm interested in?
2. When I run `DAO.new(...)`, how can I know the `invocationKey` to look for?

dev-matan-tsuberi:
@dkent600 
I mean, how do I know the `invocationKey` that is associated with a particular execution of `DAO.new(..)`, so that I know what to watch for?

```
const subscription = TransactionService.subscribe("txReceipts.DAO.new", 
  (topic, txEventInfo) => {
    if(txEventInfo.invocationKey === ???){
       // update UI
    }
});

// doesn't tell me what's the invocation key for this execution
const dao = await DAO.new(..)
```

dkent600:
>  how do I know the invocationKey that is associated with a particular execution of DAO.new(..), so that I know what to watch for?

The event passes you the parameters you passed into the invocation of `DAO.new` (if this is not clear in the docs, let me know, thanks).  Note you can include whatever you want in that object.

dev-matan-tsuberi:
@dkent600
So you say I need to include my own unique parameter in the options? like so:
```
const myRequestId = 4682648
const subscription = TransactionService.subscribe("txReceipts.DAO.new", 
  (topic, txEventInfo) => {
    if(txEventInfo.options.requestId === myRequestId){
       // update UI
    }
});

const dao = await DAO.new({
  requestId: myRequestId,
  ...
})
```
In that case, I need to manage my own unique ids and `invocationKey` is obsolete. Also I think it may break type safety since I'm adding an unknown property.

The other option is compare the `txEventInfo.options` to the options I used in the call to `DAO.new(..)`, In that case I need to worry about the defaults. And what if I make multiple calls with the same options?

docs/Home.md
```diff
@@ -101,12 +102,35 @@ await InitializeArcJs({
...
+```javascript
+await InitializeArcJs({
+  "useNetworkDefaultsFor": "kovan"
+  });
+```
+
+!!! info
+    For safety, truffle.js specifies a different HTTP port for each network.  You will need to make sure that the testnet you are using is listening on that port.  The port values are:
+
+    <table style="margin-left:2.5rem">
```
orenyodfat: Is this here by mistake ?

dkent600:
> In that case, I need to manage my own unique ids and invocationKey is obsolete. 

Correct, you would be providing your own way of tracking a specific invocation.

> Also I think it may break type safety since I'm adding an unknown property.

Not a problem.  You can extend `NewDaoConfig` however you want.

> The other option is compare the txEventInfo.options to the options I used in the call to DAO.new(..), In that case I need to worry about the defaults. And what if I make multiple calls with the same options?

This is not the greatest option, for the reasons you gave.  However, jut a note:  In the case of `DAO.new` you would not see the defaults (this is noted in the documentation).

docs/Home.md
dkent600: Do you mean the Html?  No, that is intentional

dkent600:
@orenyodfat 

Note that creating a unique Id is easy:

```
const myRequestId = Symbol("DAO.new");
```

orenyodfat:
@dkent600 I guess that the comment above are for @dev-matan-tsuberi .

docs/Home.md
dkent600: (I want the indentation)

dev-matan-tsuberi: markdown supports tables and indentation.

dkent600: > markdown supports tables and indentation.

I know, but not in the case where I needed it.  It is possible it would work in this case, but not in the original case from whence I copied this snippet.

dev-matan-tsuberi:
@dkent600 
Didn't know JavaScript had `Symbol`s, cool.

So you agree that the api as it stands, does not give a nice way to track transactions for a specific operation? (I think passing a callback is the best way for this).
Is there another use for `invocationKey` if it's cannot be used for that purpose?

Since I think this is a common use case, there should be at least an example somewhere showing how to do it using `Symbol` and extending `NewDaoConfig` as you say.

dkent600:
> So you agree that the api as it stands, does not give a nice way to track transactions for a specific operation? (I think passing a callback is the best way for this).

Depends on how you define "nice".

I am trying to avoid changing every single Arc.js function that generates a transaction, so no callback.  Providing a unique id, or whatever else you want, is little different from providing a callback, do you think?  You have control over how you do it.

BTW, note that you can use `TransactionService.generateInvocationKey` to generate your own unique invocationKey if you want.

@tibetsprague was Ok with this pattern.

> Is there another use for invocationKey if it's cannot be used for that purpose?

Yes, a UI that has a looser coupling between a specific function invocation and the UI that tracks the resulting transactions.  It is likely that Vanille will work this way.  I am not sure that Alchemy could not work this way too.

> Since I think this is a common use case, there should be at least an example somewhere showing how to do it using Symbol and extending NewDaoConfig as you say.

Done

docs/Events.md
```diff
@@ -36,7 +36,15 @@ const subscription = TransactionService.subscribe("txReceipts.DAO.new",
...
+Now you are ready to handle "txReceipts.DAO.new" events whenever you call `DAO.new`.  To let you know in advance the expected count of transactions, a "kick-off" event will be published once at the beginning of each function invocation before any transactions have begun.  In that event, `txEventInfo.tx` will be null.  The property `txEventInfo.uniqueInvocationKey` will identify the "thread" of events associated with a single function invocation.
+
+You can supply additional information in the options passed to the invoked function which are then passed back to you in the event callbacks (`txEventInfo.options`, above). For example, you may desire a tighter coupling between the events and a specific function invocation, and for you the kick-off event and invocationKey may not suffice.  In that case you could do something like:
+
+```javascript
+options.myInvocationkey = `TransactionService.generateInvocationKey("DAO.new")`;
```
dev-matan-tsuberi: rm nested backtick

dkent600: done

docs/Events.md
```diff
@@ -1,6 +1,69 @@
...
+    ```javascript
+    subscription.unsubscribe();
+    ```
+
+!!! note
+    `txEventInfo.options` will usually contain the options you passed in, with default values added.  But in the case of `DAO.new`, it will not contain the default values.  If you need the default values then instead of subscribing to "txReceipts.DAO.new" you can subscribe to "txReceipts.DaoCreator" and receive events published by  [DaoCreatorWrapper.forgeOrg](api/classes/DaoCreatorWrapper#forgeOrg) and [DaoCreatorWrapper.setSchemes](api/classes/DaoCreatorWrapper#setSchemes).  Currently this would otherwise be the same as subscribing to "txReceipts.DAO.new", though it cannot be guaranteed it will always be the case in future versions of Arc.js.
+
+### Mapping tx
+### Specifying Events
+You specify which event to which you wish to subscribe using a string which as an event specifier (or in the code, the event "topic").  The event specifier typically incorporates an Arc contract name which will map to events published by a wrapper class for that contract.  For example, the event specifier "txReceipts.DaoCreator" refers to "txReceipts" events published by [DaoCreatorWrapper](api/classes/DaoCreatorWrapper).
```
tibetsprague: "a string which as" = "a string which is"?

lib/eventService.ts
```diff
@@ -0,0 +1,122 @@
...
+   * Unsubscribe from all of the events
+   */
+  public unsubscribe(): void {
+    // timeout to allow lingering events to be handled before unsubscribing
+    setTimeout(() => {
+      this.subscriptions.forEach((s: EventSubscription) => {
+        s.unsubscribe();
+      });
+      this.subscriptions.clear();
+    }, 0);
```
tibetsprague: why does a timeout of 0 do anything?

docs/Events.md
dkent600: fixed

lib/eventService.ts
dkent600: A timeout value of 0 moves the given code to the end of the execution queue, without any further time delay, allowing everything already in the queue to execute.  In this case that would be any events already published.

Issue #176, opened 2018-04-02
Remove duplicate events


dev-matan-tsuberi:
Watching events may fire the callback twice with the same event (same tx hash). 
This is a known web3.js behaviour (https://github.com/ethereum/web3.js/issues/398).

Possible solution may be:
1. wait for a minimum number of confirmation blocks.
2. suppress duplicate events by remembering previous tx hashs (not scalable).

dkent600:
![](https://github.trello.services/images/mini-trello-icon.png) [Remove duplicate events](https://trello.com/c/CwqT1C26/120-remove-duplicate-events)

YoonjaeYoo:
@dev-matan-tsuberi When do you think it would be fixed?
We are suffering from this issue.
Thank you.

dkent600:
@dev-matan-tsuberi @YoonjaeYoo 

Have you tried this suggestion:  https://github.com/ethereum/web3.js/issues/398#issuecomment-213331868 ?

dkent600:
@YoonjaeYoo you should find that dups are now being removed in the latest release of Arc.js.  Let us know if not, thanks.

Issue #40, opened 2018-01-22
GenesisScheme wrapper should have a function that returns a collection of all the created DAOs


tibetsprague:
The main functionality is finding all the created DAOs

dkent600:
![](https://github.trello.services/images/mini-trello-icon.png) [Add function that returns a collection all DAOs](https://trello.com/c/tf1eyyQT/65-add-function-that-returns-a-collection-all-daos)

PR #197, opened 2018-04-15
Get all daos

dkent600:
Resolves: #40 
Resolves: #176 (remove duplicate events)

### DIFF
--- a/lib/contractWrapperBase.ts
+++ b/lib/contractWrapperBase.ts
@@ -192,28 +192,53 @@ export abstract class ContractWrapperBase {
     ): EventFetcher<TArgs> => {
 
       let baseEvent: EventFetcher<TArgs>;
+      let receivedEvents: Set<Hash>;
+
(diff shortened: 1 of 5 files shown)


lib/contractWrapperBase.ts
```diff
@@ -192,28 +192,53 @@ export abstract class ContractWrapperBase {
...
+        receivedEvents = new Set<Hash>();
+      }
+
+      const handleEvent = (
+        error: any,
+        log: DecodedLogEntryEvent<TArgs> | Array<DecodedLogEntryEvent<TArgs>>,
+        callback?: EventCallback<TArgs>): void => {
+
+        /**
+         * always provide an array
```
dev-matan-tsuberi: question: do these small comments appear in the generated docs? (because of the double asteriks)

dkent600: > do these small comments appear in the generated docs? (because of the double asteriks)

Not when they appear inside a function like that

Issue #144, opened 2018-03-14
log parameters sent to wrapped Arc functions


dkent600:

dkent600:
![](https://github.trello.services/images/mini-trello-icon.png) [Add logging and log levels](https://trello.com/c/RJW14vFq/19-add-logging-and-log-levels)

dkent600:
![](https://github.trello.services/images/mini-trello-icon.png) [log parameters sent to wrapped Arc functions](https://trello.com/c/uYYzzT1e/103-log-parameters-sent-to-wrapped-arc-functions)

PR #202, opened 2018-04-18
Log parameters to contract function calls

dkent600:
Resolves: #144


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@daostack/arc.js",
-  "version": "0.0.0-alpha.48",
+  "version": "0.0.0-alpha.40",
```
tibetsprague: errr

dkent600: that's weird

dkent600: i guess a typo, hit 0 instead of 9

Issue #78, opened 2018-02-08
Rename files for camel-case consistency


dkent600:
Rename files for camel-case consistency.  They are currently all over the place.

dkent600:
![](https://github.trello.services/images/mini-trello-icon.png) [Rename files for camel-case consistency](https://trello.com/c/Y88fPt0E/73-rename-files-for-camel-case-consistency)

dkent600:
All lower case, with dashes.  No camel case.  This avoids potential errors on OSs and web servers that are file name case sensitive.

Issue #58, opened 2018-01-31
Convert javascript to typescript


dkent600:
Convert javascript to typescript.  Coding would be easier and more robust.

dkent600:
#63 depends on this.

dkent600:
![](https://github.trello.services/images/mini-trello-icon.png) [Convert to TypeScript](https://trello.com/c/26XTR2S9/64-convert-to-typescript)

dkent600:
Done

PR #204, opened 2018-04-20
Converted all tests to typescript

dkent600:
Resolves: #58 
Resolves: #78 

- converts all test code to TypeScript.  This has already flushed out a few typings errors and will be great for ensuring code quality going forward.
- renames files with camel-casing
- simplifies some script names:
  - test.automated => test.run
  - test.ganache.* => ganache.* (`ganache` and `ganache.run` are synonyms)
  - test.ganacheDb.* => ganacheDb.*  (`ganacheDb` and `ganacheDb.run` are synonyms)

**Breaking**
Modified script names


lib/utils.ts
```diff
@@ -1,12 +1,16 @@
...
 import abi = require("ethereumjs-abi");
 import TruffleContract = require("truffle-contract");
-import Web3 = require("web3");
+import { providers as Web3Providers, Web3 } from "web3";
 import { gasLimitsConfig } from "../gasLimits.js";
 import { Address, DefaultSchemePermissions, Hash, SchemePermissions } from "./commonTypes";
 import { ConfigService } from "./configService";
 import { TransactionReceiptTruffle } from "./contractWrapperBase";
 import { LoggingService } from "./loggingService";
+// haven't figured out how to get web3 typings to properly expose the Web3 constructor.
```
tibetsprague: you could just do 
```
declare namespace NodeJS {
  export interface Global {
    web3: any
  }
}
```

lib/wrappers/contributionReward.ts
```diff
@@ -95,26 +95,26 @@ export class ContributionRewardWrapper extends ContractWrapperBase {
     const externalTokenReward = web3.toBigNumber(options.externalTokenReward);
 
     if (
-      (nativeTokenReward < 0) ||
-      (ethReward < 0) ||
-      (externalTokenReward < 0)
+      (nativeTokenReward.lt(0) ||
```
tibetsprague: why is this better?

dkent600: the compiler was barfing on `(externalTokenReward < 0)`, i forget exactly why.

lib/utils.ts
dkent600: The problem isn't with accessing `web3`, it is with accessing `Web3` as a constructor (note the capital 'W').  I don't want to define web3 globally anyways, as web3 is not always available to this particular code, and I don't want to risk errors occurring from any code assuming that web3 is necessarily available.

The problem here is a compiler code-generation issue that manifests at runtime -- neither the compiler nor intellisense complain, but the compiler generates the wrong code for the constructor function.  Web3 sucks with respect to how it is structured as a module and in any way that facilitates knowing the typings.  Don't suggest web3-typescript-typings because it is worthless.

For this PR I created my own web3.d.ts and they are very helpful (one for lib and one for test code).  There is a lot of redundancy between those two d.ts files, but I expect this is temporary as I am hoping that web3 1.0 will have much better support for typings out-of-the-box.

PR #214, opened 2018-05-11
Get threshold returns big number

dkent600:
Resolves: #213


lib/wrappers/genesisProtocol.ts
```diff
@@ -301,9 +301,8 @@ export class GenesisProtocolWrapper extends ContractWrapperBase implements Schem
    * The computation depends on the current number of boosted proposals in the DAO
    * as well as the GenesisProtocol parameters thresholdConstA and thresholdConstB.
    * @param {GetThresholdConfig} options
-   * @returns Promise<number>
```
dev-matan-tsuberi: `@returns Promise<BigNumber>`

dkent600: this is a removed line.  the correct value for the docs will be automatically generated.

Issue #196, opened 2018-04-13
proposingRepRewardConstB is way way way too big


tibetsprague:
Right now the default value is web3.toWei(5).This means that passing one proposal is giving a reputation reward to the proposer that looks like 50000000000000000000 even when the DAO only has 10 reputation total, and AFTER converting from WEI :) 

So, proposingRepRewardConstB needs to not be converted to WEI and also probably shouldn't be greater than 1.

tibetsprague:
https://github.com/daostack/arc/issues/438

dkent600:
@orenyodfat

PR #217, opened 2018-05-15
Mainnet migration, new GP configuration

dkent600:
This contains the new GP configuration parameters for the mainnet deploy.

### DIFF
--- a/lib/migrations/2_deploy_organization.ts
+++ b/lib/migrations/2_deploy_organization.ts
@@ -109,14 +109,10 @@ export const arcJsDeployer = (web3: Web3, artifacts: any, deployer: any): void =
       { gas: gasAmount });
 
     const AvatarInst = await Avatar.at(tx.logs[0].args._avatar);
-    const nativeTokenAddress = await AvatarInst.nativeToken();
     let genTokenAddress;
 
     if (network !== "live") {
-      // create a global GEN token
-      // await deployer.deploy(DaoToken, tokenName, tokenSymbol, 0);
-      // genTokenAddress = (await DaoToken.deployed()).address;
-      genTokenAddress = nativeTokenAddress;
+      genTokenAddress = await AvatarInst.nativeToken();;
       console.log(`using native token for staking on network != "live" at: ${genTokenAddress}`);
     } else {
       // the "real" live ETH GEN token
--- a/lib/migrations/2_deploy_organization.ts
+++ b/lib/migrations/2_deploy_organization.ts
@@ -78,8 +78,8 @@ export const arcJsDeployer = (web3: Web3, artifacts: any, deployer: any): void =
      *  Genesis DAO parameters,  FOR TESTING PURPOSES ONLY
      */
     const orgName = "Genesis Alpha";
-    const tokenName = "Gen";
-    const tokenSymbol = "GEN";
+    const tokenName = "Gdt";
+    const tokenSymbol = "Genesis Alpha";
     const orgNativeTokenFee = 0;
     const defaultVotingMachineParams = await GetDefaultGenesisProtocolParameters();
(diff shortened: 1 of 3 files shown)


dkent600:
[Deploy contracts on the main net](https://trello.com/c/4aSAr2Fv/140-deploy-contracts-on-the-main-net)

test/genesisProtocol.ts
```diff
@@ -170,8 +172,9 @@ describe("GenesisProtocol", () => {
       avatar: dao.avatar.address,
     });
     assert.isOk(result);
-    assert.equal(helpers.fromWei(result.thresholdConstA).toNumber(), 2);
-    assert.equal(result.thresholdConstB, 10);
+    assert.equal(helpers.fromWei(result.thresholdConstA).toNumber(),
```
dev-matan-tsuberi: not strictly needed, but consider using https://www.npmjs.com/package/chai-bignumber for asserting on bignumbers.

lib/wrappers/genesisProtocol.ts
```diff
@@ -1312,16 +1312,16 @@ export enum ProposalState {
 export const GetDefaultGenesisProtocolParameters = async (): Promise<GenesisProtocolParams> => {
   const web3 = await Utils.getWeb3();
   return {
-    boostedVotePeriodLimit: 604800, // 1 week
+    boostedVotePeriodLimit: 604800,
```
dev-matan-tsuberi: should be 3 days (259200) instead of a week, no?

lib/migrations/2_deploy_organization.ts
```diff
@@ -110,12 +110,25 @@ export const arcJsDeployer = (web3: Web3, artifacts: any, deployer: any): void =
...
+
+    if (network !== "live") {
+      // create a global GEN token
+      // await deployer.deploy(DaoToken, tokenName, tokenSymbol, 0);
+      // genTokenAddress = (await DaoToken.deployed()).address;
+      genTokenAddress = nativeTokenAddress;
+      console.log(`using global GEN token for network != "live" at: ${genTokenAddress}`);
+    } else {
+      // the "real" live ETH GEN token
+      genTokenAddress = "0x543Ff227F64Aa17eA132Bf9886cAb5DB55DCAddf";
```
tibetsprague: how come this address isn't the Avatar's nativeToken()? shouldn't it be?

lib/wrappers/genesisProtocol.ts
```diff
@@ -910,62 +910,62 @@ export interface RedeemEventResult {
...
-   * Default is 5, converted to Wei.
+   * Default is 50, converted to Wei.
    */
   proposingRepRewardConstA: BigNumber.BigNumber | string;
   /**
    * Constant B in the calculation of the proposer's reward.
    * See [[GenesisProtocolWrapper.getRedeemableReputationProposer]].
    * Must be between 0 and 100.
-   * Default is 1.
+   * Default is 50.
```
tibetsprague: im a bit concerned that this is too big. remember https://github.com/daostack/arc.js/issues/196? arc may need to change to treat it as a percentage right? we will test though. @orenyodfat @fmatan

lib/migrations/2_deploy_organization.ts
tibetsprague: it is? how is it saying that? it seems to me that GEN should always be the native token of the Genesis DAO when deployed on any network

dev-matan-tsuberi: this gets passed to the constructor of `GenesisProtocol` which accepts the staking token.

dev-matan-tsuberi: https://github.com/daostack/arc/blob/master/contracts/VotingMachines/GenesisProtocol.sol#L96

PR #218, opened 2018-05-16
Alchemy mainnet deployment

orenyodfat:
- Use truffle-hdwallet-provider
- Add migrate scripts for alchemy needed contracts.
- In order to deploy to main net one need to modify config/secrets.json adding its own 12 words mnemonic and the node provider url.
- config/secrets.json added to .gitignore


config/secrets.json
```diff
@@ -0,0 +1,4 @@
+{
```
dkent600: This file should be supplied in the same manner as the founders json file, via environment variable that specifies the path.

lib/migrations/2_deploy_alchemy.ts
```diff
@@ -0,0 +1,190 @@
+import { Web3 } from "web3";
```
dkent600: There is currently no anticipated use for this file so I think it should be removed from the PR.  We can't publish Arc.js to NPM in a way that only works for Alchemy.

lib/migrations/2_deploy_organization.ts_
```diff
@@ -0,0 +1,237 @@
+import { Web3 } from "web3";
```
dkent600: I'm not clear why this file is here.  ".ts_"?

dkent600:
@orenyodfat Can you please add a comment somewhere in the code that explains why we need to use this provider?  Specifically, yhy it enables the mainnet deploy to succeed?

config/secrets.json
dkent600: I can code that if you wish

PR #223, opened 2018-05-18
New Genesis Protocol bounty API

dkent600:
- added to `GenesisProtocol`:
  - `getTokenBalances`
  - `getRedeemableTokensStakerBounty`
  - `redeemDaoBounty`
  - `RedeemDaoBounty` (new event)

- added tests for the above
- cleaned up some redundant `GenesisProtocol` and `ContributionReward` interfaces
- cleaned up and documented new migration methodology


docs/Migration.md
```diff
@@ -4,35 +4,37 @@
...
-4. Make sure you have your testnet running and listening on the right port. 
+    ```
+    {
+      "mnemonic" : "a 12-word BIP39-compliant account mnemonic",
+      "providerUrl" : "examples: http://127.0.0.1:8545 or https://[network].infura.io/[token]"
+    }
+    ```
 
-      For safety, truffle.js specifies a different HTTP port for each network.  You will need to make sure that the testnet you are using is listening on that port.  The port values are:
+    An account generated from the mnemonic will be locked for signing and accepting the transactions that are generated during the migration.
```
tibetsprague: unlocked right?

lib/migrations/2_deploy_organization.ts
```diff
@@ -29,12 +30,12 @@ export const arcJsDeployer = (web3: Web3, artifacts: any, deployer: any): void =
   // so Utils.getWeb3 can find it
   (global as any).web3 = web3;
 
-  const network = ConfigService.get("network") || "ganache";
+  const network = env.arcjs_network || "ganache";
```
tibetsprague: why this change?

dkent600: I can't think of a use case for enabling the migration-related env variables to be defined in the config system.  It adds needless complication.  So I removed it.  Let me know if you can think of a use case.

docs/Migration.md
dkent600: oops

PR #232, opened 2018-05-25
fix getSchemePermissions

dkent600:
Fixes bug in `getSchemePermissions` and adds a couple other minor enhancements.

### DIFF
--- a/lib/contractWrapperBase.ts
+++ b/lib/contractWrapperBase.ts
@@ -168,7 +168,7 @@ export abstract class ContractWrapperBase {
    * @param avatarAddress
    */
   protected async _getSchemePermissions(avatarAddress: Address): Promise<SchemePermissions> {
-    const permissions = (await this.getController(avatarAddress))
+    const permissions = await (await this.getController(avatarAddress))
       .getSchemePermissions(this.address, avatarAddress) as string;
 
     return SchemePermissions.fromString(permissions);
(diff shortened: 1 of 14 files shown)


lib/contractWrapperBase.ts
```diff
@@ -168,7 +168,7 @@ export abstract class ContractWrapperBase {
    * @param avatarAddress
    */
   protected async _getSchemePermissions(avatarAddress: Address): Promise<SchemePermissions> {
-    const permissions = (await this.getController(avatarAddress))
+    const permissions = await (await this.getController(avatarAddress))
```
dev-matan-tsuberi: why the double `await`?

dkent600: there are two async functions being called here.  the bug is that I wasn't awaiting the second one

dev-matan-tsuberi: But when you return a promise from a `then` callback, it flattens it into a single promise. Ther'e can't be a promise of a promise (`Promise<Promise<T>>`).

dkent600: The parens ensure that it is not a promise of a promise.  See also the test I added that confirms that the function works now.

dev-matan-tsuberi: single `await` should work just as well as double `await`.

dkent600: It doesn't because you end up getting a promise from the second async call instead of the actual value being promised.

PR #234, opened 2018-05-29
handle logging circular json

dkent600:
Fix crashes when stringifying objects having circular references.

### DIFF
--- a/lib/loggingService.ts
+++ b/lib/loggingService.ts
@@ -1,4 +1,5 @@
 import { ConfigService } from "./configService";
+import * as JSON from 'circular-json';
 
 export enum LogLevel {
   none = 0,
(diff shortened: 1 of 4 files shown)


dkent600:
@fmatan 

> I don't think it's good to expose a different unconventional JSON format.

The altered JSON object is scoped within LoggingService and is only used for stringifying objects for the purpose of logging.  Logging output doesn't need to be JSON at all, it can be whatever we want it to be, I just happen to be using JSON for ease of coding.

dev-matan-tsuberi:
@dkent600 Still users need to learn an unfamiliar JSON format to understand the logs. I think it's better to use plain flat objects. (There's also the risk that you try to serialize an object with references all over the place in which case you end up with essentially a huge dump of what's in memory)

Issue #243, opened 2018-06-06
GenesisProtocol Wrapper: getter for the staking token.


dev-matan-tsuberi:
A getter for the staking token is missing.

dkent600:
@dev-matan-tsuberi Coincidentally I already added this in a branch I'm working right now.  In the meantime you can get the address from `await genesisProtocolWrapper.contract.stakingToken()`.

Issue #241, opened 2018-06-04
Use latest truffle version during migrations


dkent600:
Use latest truffle version during migrations, at least to mainnet, while maintaining the functionality of truffle-core-migrate-without-compile in the context of applications.

Options: 
- check whether running in repo (by existence of .git folder?) and always use global truffle command in that case (will we be happy with truffle's behavior wrt recompiling?).
- find a way to keep truffle-core-migrate-without-compile up to date with the latest truffle releases.

dkent600:
@orenyodfat

Issue #239, opened 2018-06-04
Separate deployment of the Genesis DAO in the Arc.js deployment script


dkent600:
Deploy the Genesis DAO in a separate deployment script, or perhaps better, make it an optional part of the deployment script.

Applications will still want to be able to deploy both, at least to ganache, so there should remain a single script that does both.

dkent600:
@orenyodfat

orenyodfat:
better to have it in two separate scripts.
first script deploy : 
- Schemes 
- VotingMachines 
- DaoCreator

second script create dao and register schemes: 
- get a set of scheme (which already been deployed at the first script).
- dao params (token name, symbol,founders....).

user may choose to avoid using the second script .

dkent600:
![](https://github.trello.services/images/mini-trello-icon.png) [Separate deployment of the Genesis DAO in the Arc.js deployment script](https://trello.com/c/EQ1e0zoW/144-separate-deployment-of-the-genesis-dao-in-the-arcjs-deployment-script)

PR #244, opened 2018-06-06
Split migration script

dkent600:
Resolves: #243 
Resolves: #241
Resolves: #239 

- In hopes of improving the prospects when deploying to the mainnet, and in anticipation that Genesis DAO creation may no longer be part of the Arc.js migration, splits deployment script into separate parts:

1. the deployment script that deploys everything (`npm run migrateContracts`), and 
2. a script that creates the Genesis DAO (`npm run createGenesisDao`).

- cleans up code in package-scripts
- adds `getStakingToken` function to `GenesisProtocolWrapper`
- uses global truffle command when running in the context of a repository (looks for ".git" folder)


lib/migrations/3_deploy_schemes.ts
```diff
@@ -0,0 +1,70 @@
...
+    console.log(`Deploying schemes to ${network}, gasLimit: ${gasLimit}`);
+
+    let genTokenAddress;
+
+    if ((network !== "live") && (network !== "kovan")) {
+      genTokenAddress = migrationState.nativeTokenAddress;
+      console.log(`Using Genesis native token from ${migrationState.orgName} for staking on network ${network} at: ${genTokenAddress}`);
+    } else {
+      // the global GEN token
+      genTokenAddress = "0x543Ff227F64Aa17eA132Bf9886cAb5DB55DCAddf";
```
tibetsprague: im confused, you are using the same token for Kovan and live?

dkent600: Yes, GEN is deployed to the same address in both places.

orenyodfat:
How about arc.js migration scripts will  deploy only the schemes,UController,DAOCreator and VotingMachines ?
It will also provide an api to forgeOrg .(not in the deployment script).

dkent600:
@orenyodfat 

> It will also provide an api to forgeOrg .(not in the deployment script).

Can you please clarify?

dkent600:
@orenyodfat I think I have addressed your suggestion above.  Basically there are two parts:  1) the deployment script that deploys everything (`npm run migrateContracts`), and 2) a script that creates the Genesis DAO (`npm run createGenesisDao`).There is an important condition for enabling this to happen:  When we are deploying the schemes to a network that has no GEN token at the standard address (currently the case with ganache and ropsten), then, when the network is ganache we create a GEN token and mint it to all the accounts so tests will succeed;  but when the network is ropsten we throw an exception.

In other words with this model we can't migrate to ropsten until we deploy GEN to it.

I will make the script treat ropsten like ganache iff you give a thumbs up that it would be OK.

gasLimits.js
```diff
@@ -21,17 +21,35 @@ const gasLimitsConfig =
...
 
+/**
+ * Returns promise of the maximum gasLimit that we dare to ever use, given the
+ * current state of the chain.
+ * @param {} web3 
+ */
+const computeMaxGasLimit = function (web3) {
+  return promisify((callback) => web3.eth.getBlock("latest", false, callback))()
+    .then((block) => {
+      return block.gasLimit - 50000;
```
orenyodfat: I would suggest to replace 50000 to 100000 just to be on the safe side

lib/migrations/createGenesisDao.ts
```diff
@@ -1,10 +1,16 @@
-import { promisify } from "es6-promisify";
 import { Web3 } from "web3";
```
orenyodfat: How about moving this script out of the **migration** folder to something like **scripts** folder ?

lib/migrations/createGenesisDao.ts
```diff
@@ -23,135 +29,130 @@ interface FounderSpec {
...
-   */
-  deployer.then(async () => {
+    const live = this.network === "live";
     /**
-     *  Genesis DAO parameters,  FOR TESTING PURPOSES ONLY
+     * Genesis DAO parameters
      */
-    const orgName = "Genesis Alpha";
-    const tokenName = "Genesis Alpha";
+    const orgName = live ? "Genesis" : "Genesis Alpha";
```
orenyodfat: Why "Genesis" ? Why "GenesisAlpha"  Alpha ?
How about rename this script to createDao and pass to it a params (
e.g orgName... ).
Then this script can be used to deploy any Dao .

lib/migrations/createGenesisDao.ts
```diff
@@ -23,135 +29,130 @@ interface FounderSpec {
...
-  const SchemeRegistrar = artifacts.require("SchemeRegistrar.sol");
-  const SimpleICO = artifacts.require("SimpleICO.sol");
-  const TokenCapGC = artifacts.require("TokenCapGC.sol");
-  const UController = artifacts.require("UController.sol");
-  const UpgradeScheme = artifacts.require("UpgradeScheme.sol");
-  const VestingScheme = artifacts.require("VestingScheme.sol");
-  const VoteInOrganizationScheme = artifacts.require("VoteInOrganizationScheme.sol");
-  const GenesisProtocol = artifacts.require("GenesisProtocol.sol");
-  const ControllerCreator = artifacts.require("ControllerCreator.sol");
+  public async forge(foundersConfigurationLocation: string): Promise<ForgedDaoInfo> {
```
orenyodfat: How is calling this function ?
How about adding module.exports.createDao =function(params...) {
...do the dao creation...
}
then it could be easily be executed externaly from the command line with something like : 
`node -e 'require("./createDao.js").createDao(params..))'`

lib/migrations/createGenesisDao.ts
```diff
@@ -23,135 +29,130 @@ interface FounderSpec {
...
+      { gas: gasLimit });
 
-    // maximize chances that Avatar.at will succeed
-    if (network === "live") {
-      await UtilsInternal.sleep(60000);
+    /**
+     * save info for later steps
+     */
+    if (live) {
+      UtilsInternal.sleep(10000); // maximize chances this is ready
```
orenyodfat: Does this sleep is a must ?
I am not sure.
We can use a better way to confirm all is ready such as checking it..

lib/migrations/createGenesisDao.ts
```diff
@@ -23,135 +29,130 @@ interface FounderSpec {
...
-    }
     /**
-     * The voting machine.  GenesisProtocol must be deployed as a scheme if it is
-     * to be used by schemes as a voting machine, which is what all of the
-     * Genesis schemes do.
+     * Truffle Solidity artifact wrappers
      */
-    await deployer.deploy(GenesisProtocol, genTokenAddress, { gas: gasAmount });
-    const genesisProtocolInst = await GenesisProtocol.deployed();
+    const ContributionReward = await Utils.requireContract("ContributionReward");
```
orenyodfat: The schemes list suppose to be passed from the user unless it choose the defaults.

package-scripts.js
```diff
@@ -10,43 +10,44 @@ const env = require("env-variable")();
...
-
-const ganacheCommand = `ganache-cli -l ${computeGasLimit(6)} --account="0x8d4408014d165ec69d8cc9f091d8f4578ac5564f376f21887e98a6d33a6e3549,9999999999999999999999999999999999999999999" --account="0x2215f0a41dd3bb93f03049514949aaafcf136e6965f4a066d6bf42cc9f75a106,9999999999999999999999999999999999999999999" --account="0x6695c8ef58fecfc7410bf8b80c17319eaaca8b9481cc9c682fd5da116f20ef05,9999999999999999999999999999999999999999999" --account="0xb9a8635b40a60ad5b78706d4ede244ddf934dc873262449b473076de0c1e2959,9999999999999999999999999999999999999999999" --account="0x55887c2c6107237ac3b50fb17d9ff7313cad67757e44d1be5eb7bbf9fc9ca2ea,9999999999999999999999999999999999999999999" --account="0xb16a587ad59c2b3a3f47679ed2df348d6828a3bb5c6bb3797a1d5a567ce823cb,9999999999999999999999999999999999999999999"`;
-const ganacheDbCommand = `ganache-cli --db ${pathDaostackArcGanacheDb} -l ${computeGasLimit(6)} --networkId 1512051714758 --mnemonic "behave pipe turkey animal voyage dial relief menu blush match jeans general"`;
+const gasLimit = 8000000; // something reasonably close to live
```
orenyodfat: why is that ?

lib/migrations/createGenesisDao.ts
dkent600: `forge` and `setSchemes` are called from /package-scripts/createGenesisDao.js which is invoked from package-scripts.js.

Regarding module.exports, we already have the `DAO` class on which applications can call `DAO.new` to create a DAO given a configuration object.

I am not using `DAO.new` in the migration script mainly I want to minimize the complexity of the script, and there has been no request to support the ability to receive arbitrary DAO configurations, other than Genesis.

In any case, I would prefer to use a separate PR for creating a script to support adding arbitrary DAOs.

package-scripts.js
dkent600: It is only used for ganache

gasLimits.js
dkent600: Done

lib/migrations/createGenesisDao.ts
dkent600: Done

lib/migrations/createGenesisDao.ts
dkent600: Seems like a reasonable feature request to me, but I would prefer to make that a separate PR,  Is Ok with you?

lib/migrations/createGenesisDao.ts
dkent600: Refactored to only sleep when necessary

lib/migrations/createGenesisDao.ts
dkent600: Separate PR to support creating arbitrary DAOs.

gasLimits.js
```diff
@@ -1,37 +1,55 @@
...
-    /**
-     * Gas limit appropriate for all other transactions besides creating DAOs
-     */
-    "gasLimit_runtime": 4543760,
-  };
+{
+  /**
+   * The gas limit used by Arc to forge a DAO with foundersInGasLimitArc founders
+   */
+  "gasLimit_arc": 5600000,
```
orenyodfat: this value can and should be taken from arc
(e.g. test/constant)

lib/migrations/2_deploy_schemes.ts
```diff
@@ -0,0 +1,98 @@
...
+    let genTokenAddress;
+
+    await DAOToken.at("0x543Ff227F64Aa17eA132Bf9886cAb5DB55DCAddf")
+      .then((token: any): void => genTokenAddress = token.address)
+      .catch(() => { console.log(`global GEN token does not exist`); });
+
+    if (!genTokenAddress) {
+      if (network === "ganache") {
+        // then we will create a new token to use for staking
+        const genToken = await DAOToken.new("GEN", "GEN", web3.toWei(100000000));
```
orenyodfat: why calling that GEN ?

lib/migrations/2_deploy_schemes.ts
```diff
@@ -0,0 +1,98 @@
...
+        genTokenAddress = genToken.address;
+        // and mint GEN to all of the accounts
+        const getAccounts = promisify(web3.eth.getAccounts);
+        const accounts = await getAccounts();
+        accounts.forEach((account: Address) => {
+          genToken.mint(account, web3.toWei(1000));
```
orenyodfat: from where come the 1000 ?

lib/migrations/2_deploy_schemes.ts
```diff
@@ -0,0 +1,98 @@
...
+    await deployer.deploy(UpgradeScheme, { gas: gasLimit });
+    await deployer.deploy(GlobalConstraintRegistrar, { gas: gasLimit });
+    await deployer.deploy(ContributionReward, { gas: gasLimit });
+    await deployer.deploy(AbsoluteVote, { gas: gasLimit });
+    await deployer.deploy(SimpleICO, { gas: gasLimit });
+    await deployer.deploy(TokenCapGC, { gas: gasLimit });
+    await deployer.deploy(VestingScheme, { gas: gasLimit });
+    await deployer.deploy(VoteInOrganizationScheme, { gas: gasLimit });
+    if (network !== "live") {
+      await deployer.deploy(ExecutableTest, { gas: gasLimit });
```
orenyodfat: why ExecutableTest need to be part of the deployment ? 
who is using that ?

lib/migrations/createGenesisDao.ts
orenyodfat: why separate PR ?
This should be in this pr.

orenyodfat: Why "GenesisAlpha" Alpha ?
It is just another DAO

lib/migrations/createGenesisDao.ts
orenyodfat: If the script written in a good way it should not use  sleep at all...

lib/scripts/createGenesisDao.ts
```diff
@@ -23,135 +29,132 @@ interface FounderSpec {
...
-      await UtilsInternal.sleep(60000);
+    let avatarInst;
+
+    /**
+     * save info for later steps
+     */
+    while (!avatarInst) {
+      avatarInst = await Avatar.at(txForgeOrg.logs[0].args._avatar);
+      console.log("sleeping until Avatar is mined...");
+      UtilsInternal.sleep(10000); // sleep and retry until avatarInst is ready
```
orenyodfat: I guess that 10 seconds sleep is too much here..
You can sleep for 1 second.
And add a comment that it is "not nice hack".

gasLimits.js
dkent600: Arc is a _dev_ dependency of Arc.js, so Arc can't be expected to be present in an application context.  Thus we manually copy this value from Arc.

lib/migrations/2_deploy_schemes.ts
dkent600: That is just a number we've always used for the ganache founders

lib/migrations/2_deploy_schemes.ts
dkent600: There is an Arc.js test that uses it.

gasLimits.js
orenyodfat: though you are importing arc .

dkent600: We aren't literally "importing" Arc.  But we do have a dev _dependency_ on it so we can grab the contract json files at contract migration time.

Because it is a dev dependency, applications will not pull in the Arc package.  There is no need for them to do so.

The only time this value is used is when we compute the amount of gas need to forge an org.  I have been considering adding the ability for applications to customize this gas limit via the arc.js configuration options, but thus far I've heard no need for it.

lib/migrations/createGenesisDao.ts
dkent600: We still don't know why it was needed in the first place. Until we understand, I would prefer to leave it in.  I have commented the code accordingly.

PR #251, opened 2018-06-20
Imports Arc gas limit from Arc, instead of hard-coding

dkent600:
Imports Arc gas limit from Arc, instead of hard-coding

### DIFF
--- a/package-scripts/importArcGasLimit.js
+++ b/package-scripts/importArcGasLimit.js
@@ -1,0 +1,6 @@
+const fs = require("fs-extra");
+
+const arcPackage = require("../node_modules/@daostack/arc/package.json");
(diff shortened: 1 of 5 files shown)


package-scripts/importArcGasLimit.js
```diff
@@ -0,0 +1,6 @@
+const fs = require("fs-extra");
+
+const arcPackage = require("../node_modules/@daostack/arc/package.json");
+const gasLimit = arcPackage.config.gasLimit;
+
+fs.writeFileSync("./arcConstants.js", `const ARC_GAS_LIMIT = ${gasLimit};\r\nmodule.exports = { ARC_GAS_LIMIT };\r\n`);
```
orenyodfat: how about just copy arc gas limit file as it ?

PR #263, opened 2018-06-28
Adds StandardToken.approve wrapper

dkent600:
Adds `StandardToken.approve` wrapper

### DIFF
--- a/lib/iContractWrapperBase.ts
+++ b/lib/iContractWrapperBase.ts
@@ -47,6 +47,9 @@ export class ArcTransactionResult {
    * Returns null if the transaciton is not yet mined.
    */
   public async getTxMined(): Promise<TransactionReceiptTruffle | null> {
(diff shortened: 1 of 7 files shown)


lib/iContractWrapperBase.ts
```diff
@@ -47,6 +47,9 @@ export class ArcTransactionResult {
    * Returns null if the transaciton is not yet mined.
    */
   public async getTxMined(): Promise<TransactionReceiptTruffle | null> {
+    if (!this.tx) {
```
dev-matan-tsuberi: why?

dkent600: Because StandardToken.approve returns an ArcTransactionResult with tx === null when the amount to approve === 0.  I've considered using this elsewhere as a means of indicating that no tx has been sent.

PR #270, opened 2018-07-05
remove TxEventContext from payload

dkent600:
Removes `TxEventContext` from payload to avoid problems for the client, such as circular references to `options`.

### DIFF
--- a/lib/transactionService.ts
+++ b/lib/transactionService.ts
@@ -132,9 +132,9 @@ export class TransactionService extends PubSubEventService {
    * Will take obj.txEventContext, else create a new one.
    *
    * @hidden - for internal use only
+   * @param functionName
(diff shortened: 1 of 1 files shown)


lib/transactionService.ts
```diff
@@ -553,10 +553,16 @@ export class TransactionService extends PubSubEventService {
       payload.tx = tx;
       payload.txReceipt = txReceipt;
       payload.txStage = txStage;
+
+      // delete so doesn't cause problems (such as circular references to `options`) for the client
+      const savetxEventContext = payload.options.txEventContext;
+      payload.options.txEventContext = undefined;
```
dev-matan-tsuberi: setting to `undefined` is not the same as `delete`ing

PR #278, opened 2018-07-10
Upgrade to arc44

dkent600:
- New Redeemer contract for redeeming from ContributionRewards and GenesisProtocol in a single tx.  It is a full-fledged deployed wrapper with methods `redeem` and `getRedemptions`
- New `GenesisProtocol.stakeWithApproval` method that pre-approves staking token transfer and does the stake in a single tx
- `GenesisProtocol.ExecuteProposal` event loses _executionState param
- new event: `GenesisProtocol.GPExecuteProposal`
- New event `AbsoluteVote.RefreshReputation`
- `AbsoluteVote.VoteProposal` event loses `_ownerVote` param 
- new event: `AbsoluteVote.AVVoteProposal`
- `IntVoteInterface` now exports the full suite of events
- Indexed `_avatar` added to all `IntVoteInterface` events
- `AgreementProposal` event now has indexed `proposalId`
- New `ProposeVestedAgreement` event
- In `GenesisProtocol.Stake` event:  "_voter" renamed to "_staker"
- Greater precision in `GenesisProtocol` threshold calculation
- `VestingScheme.AgreementProposal._proposalId` is now indexed
- New `getAllowedRangeOfChoices` on IntVoteInterface
- New `getGlobalConstraintParameters` on (U)Controller.  `TokenCapGC` now has the ability to report on its parameters.
- GP threshold computation now excludes expired boosted proposals
- `QuorumVote` and `OrganizationRegister` are now being deployed
- removed wrapper support for voting and staking on behalf of another account
- upgrades ganache-cli version

**Breaking**
- `GenesisProtocol.ExecuteProposal` event loses `_executionState` param
- `AbsoluteVote.VoteProposal` event loses `_ownerVote` param 
- In `GenesisProtocol.Stake` event:  "_voter" renamed to "_staker"
-  removed wrapper support for voting and staking on behalf of another account
- requires upgrade to ganache-cli version 6.1.6

### DIFF
--- a/lib/wrappers/genesisProtocol.ts
+++ b/lib/wrappers/genesisProtocol.ts
@@ -1,5 +1,6 @@
 "use strict";
 import * as BigNumber from "bignumber.js";
+import { gasLimitsConfig } from "../../gasLimits.js";
 import { AvatarService } from "../avatarService";
 import {
   Address,
@@ -23,14 +24,13 @@ import { Utils } from "../utils";
 import { EntityFetcherFactory, EventFetcherFactory, Web3EventService } from "../web3EventService";
 import { RedeemEventResult } from "./commonEventInterfaces";
 import {
+  ExecuteProposalEventResult,
   NewProposalEventResult,
   OwnerVoteOptions,
   ProposalIdOption,
   ProposeOptions,
   VoteOptions,
-  VoteProposalEventResult,
   VoteWithSpecifiedAmountsOptions,
-  VotingMachineExecuteProposalEventResult,
 } from "./iIntVoteInterface";
 
 import { IntVoteInterfaceWrapper } from "./intVoteInterface";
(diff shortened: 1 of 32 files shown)


lib/wrappers/genesisProtocol.ts
```diff
@@ -67,20 +67,13 @@ export class GenesisProtocolWrapper extends IntVoteInterfaceWrapper implements S
   public async stake(options: StakeConfig =
     {} as StakeConfig & TxGeneratingFunctionOptions): Promise<ArcTransactionResult> {
 
-    const defaults = {
-      onBehalfOf: null,
```
tibetsprague: im actually not sure we should remove this from arc.js, if it exists in arc we should probably support it, even if im worried about the feature in general

dkent600: I felt that our concerns about it were significant enough to warrant erring on the cautious side.

PR #284, opened 2018-07-17
Migration tweaks

dkent600:
- supply gas price to individual deployments
- make web3.eth.sign not synchronous
- expose Redeemer typings
- includes mainnet contract addresses

### DIFF
--- a/lib/scripts/createGenesisDao.ts
+++ b/lib/scripts/createGenesisDao.ts
@@ -55,8 +55,8 @@ export class GenesisDaoCreator {
     /**
      * Genesis DAO parameters
      */
-    const orgName = live ? "Genesis" : "Genesis Alpha";
-    const tokenName = live ? "Genesis" : "Genesis Alpha";
+    const orgName = live ? "Genesis Alpha" : "Genesis Alpha";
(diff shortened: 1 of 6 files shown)


lib/scripts/createGenesisDao.ts
```diff
@@ -55,8 +55,8 @@ export class GenesisDaoCreator {
     /**
      * Genesis DAO parameters
      */
-    const orgName = live ? "Genesis" : "Genesis Alpha";
-    const tokenName = live ? "Genesis" : "Genesis Alpha";
+    const orgName = live ? "Genesis Alpha" : "Genesis Alpha";
```
tibetsprague: heh

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@daostack/arc.js",
-  "version": "0.0.0-alpha.64",
+  "version": "0.0.0-alpha.65",
```
tibetsprague: wait aren't we already at 65?

lib/scripts/createGenesisDao.ts
dkent600: just ignore that  ;-)

PR #288, opened 2018-07-22
fix bugs in createGenesisDao against mainnet

dkent600:
fixes bugs in createGenesisDao against mainnet in the code that waits for the avatar to be accessible

### DIFF
--- a/lib/scripts/createGenesisDao.ts
+++ b/lib/scripts/createGenesisDao.ts
@@ -110,18 +110,24 @@ export class GenesisDaoCreator {
 
     let avatarInst;
 
-    /**
-     * save info for later steps
-     */
     while (!avatarInst) {
-      avatarInst = await Avatar.at(txForgeOrg.logs[0].args._avatar);
-      console.log("sleeping until Avatar is mined...");
-      /**
-       * Sleep and retry until avatarInst is ready.
-       * This is an unfortunate hack until we better understand
-       * why it has been needed when deploying to mainnet.
-       */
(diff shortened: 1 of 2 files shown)


lib/scripts/createGenesisDao.ts
```diff
@@ -110,18 +110,24 @@ export class GenesisDaoCreator {
 
     let avatarInst;
 
-    /**
-     * save info for later steps
-     */
     while (!avatarInst) {
```
dev-matan-tsuberi: Can't you just wait using `TransactionService`?

dkent600: I thought about that.  Thing is, I'm not sure why it is necessary in the first place.  Truffle normally waits until something is mined before returning from the promise.  So why is the avatar not available?  I don't know, but I don't think TxService.watchForMined would help.

dkent600: Just a note:  I've generally been avoiding using Arc.js services unless really needed, just to keep things simple and clear in the deployment script.  In this case I'd feel justified in using it, but I'm not convinced it would help, and didn't feel it was important enough to spend $$$ on testing one way or the other, since this code works.

dev-matan-tsuberi: What was the problem with the original code?

dkent600: was missing an `await` and was not catching the exception when thrown by `.at`

PR #294, opened 2018-07-26
Add the ability to set required depth on watch and subscribe

dkent600:
Adds the ability to set required depth on `watch` and `subscribe` on wrapper event and entity fetchers.

### DIFF
--- a/lib/web3EventService.ts
+++ b/lib/web3EventService.ts
@@ -103,7 +103,8 @@ export class Web3EventService {
 
         subscribe(
           eventName: string,
-          callback?: EventWatchSubscriptionCallback<TEventArgs>): Web3EventSubscription<TEventArgs> {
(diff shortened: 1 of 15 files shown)


lib/web3EventService.ts
```diff
@@ -364,8 +342,19 @@ export class Web3EventService {
...
-          log.forEach((e: DecodedLogEntryEvent<TEventArgs>) => { callback(error, e); });
+          for (const e of log) {
+            if (requiredDepth) {
+              await TransactionService.watchForConfirmedTransaction(e.transactionHash, null, requiredDepth);
+            }
+            callback(error, e);
+          }
         } else {
+
+          for (const e of log) {
```
tibetsprague: isn't this duplicate code of above?

dkent600: no, but it is redundant.  I've cleaned it up.  Good catch.

PR #301, opened 2018-07-30
GenesisProtocol : support eth_signTypedData with Metamask

orenyodfat:
depend on daostack/arc#507


dkent600:
This change is done here: https://github.com/daostack/arc.js/pull/305

PR #305, opened 2018-08-01
Upgrade to Arc47

dkent600:
This is straight 47, including the new `Redeemer` with no more event.

- `GenesisProtocolWrapper.getThreshold` no longer takes `proposalId` as a parameter
- the `Redeemer.Redeem` event has been removed
- `GenesisProtocolWrapper.stakeWithApproval` works in Metamask now

**Breaking**
- the `Redeemer.Redeem` event has been removed
- `genesisProtocol.getThreshold` no longer takes `proposalId` as a parameter

### DIFF
--- a/lib/transactionService.ts
+++ b/lib/transactionService.ts
@@ -507,6 +507,24 @@ export class TransactionService extends PubSubEventService {
     return result;
   }
 
+  /**
+   * Returns the default value for required block depth defined for the current network
+   * in the Arc.js global configuration ("txDepthRequiredForConfirmation").
+   * @param requiredDepth Overrides the default if given
+   */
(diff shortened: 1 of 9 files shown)


lib/transactionService.ts
```diff
@@ -507,6 +507,24 @@ export class TransactionService extends PubSubEventService {
...
   }
 
+  /**
+   * Returns the default value for required block depth defined for the current network
+   * in the Arc.js global configuration ("txDepthRequiredForConfirmation").
+   * @param requiredDepth Overrides the default if given
+   */
+  public static async getDefaultDepth(requiredDepth?: number): Promise<number> {
+
+    if (typeof requiredDepth === "undefined") {
```
tibetsprague: why would you pass in a requiredDepth and then just return that?

Issue #315, opened 2018-08-03
LogLevel should work via config setting


dkent600:
LogLevel should work via the ConfigService setting.  It doesn't.  It only works via LoggingService.logLevel.

PR #327, opened 2018-08-10
update logLevel  on config changes

dkent600:
Resolves: #315


test/utils.ts
```diff
@@ -11,6 +11,13 @@ import * as helpers from "./helpers";
 
 describe("Misc", () => {
 
+  it("can update logLeve via config", async () => {
```
tibetsprague: missing L

PR #346, opened 2018-08-30
Fixed a bug in `getCurrentVoteStatus`

ben-kaufman:
Changed loop condition from `choice < numChoices` to `choice <= numChoices` to include the full range of choices.


ben-kaufman:
@dkent600 sure, when I want to know the votes on a proposal currently I'll get all votes options except the last one. Meaning if for example the choices are either 0 (abstain) 1 (yes) 2 (no) then I'll only get an array of 2 with the votes for the abstain and yes without getting the no votes.

ben-kaufman:
This is simply a bug fix not a new feature.

dkent600:
@ben-kaufman OK I think I get it, if the voting machine is AbsoluteVote, then `abstainAllow` is true and the array length will be 2.

I wonder this is actually a bug in Arc that AbsoluteVote will return '2' rather than '3' (ie, it won't include abstain).  At any rate, there is some confusion here in how the abstain voting option is being treated, either in Arc or Arc.js.

ben-kaufman:
Yes I agree the problem is that abstain is mistreated either by Arc or Arc.js.
It is also very confusing to understand if 0 is yes/no/abstain because of that.
@orenyodfat what do you think?

ben-kaufman:
I think this change now should make it work for both cases, however, I still think we'll need to discuss this logic.

ben-kaufman:
@dkent600 could you please check out if the new fix works for you as well?

PR #348, opened 2018-09-05
use AbsoluteVote for schemes in the createGenesisDao script

dkent600:
Use AbsoluteVote for schemes in the createGenesisDao script.

This script is not used for creating the DAOs in use by DAOstack/Alchemy, rather is intended as a quick-and-dirty way to simulate the Genesis DAO in test environments.  So this change is not terribly important, however, but does keep the resulting test DAO consistent with what we're doing with the real Genesis DAO.


dkent600:
Closing to completely rewrite the script in a different PR

Issue #342, opened 2018-08-24
tweak formula for computing forgeOrg gas.  Cap at block limit


dkent600:
Tweak formula for computing forgeOrg gas according to the number of founders -- it is currently computing too high.  Cap the computation at the current block gas limit -- with 54 founders it is exceeding the limit.

dkent600:
My test results:

gasEstimate with one founder: 2272318
gasEstimate with two founders: 2307606
cost of additional founder: 35288

dkent600:
When using non-universal controller:

gasEstimate with one founder: 5481352
gasEstimate with two founders: 5516640
cost of additional founder: 35288

dkent600:
gasEstimate with three founders: 5551800

dkent600:
Looks like using `estimateGas` will suffice

PR #349, opened 2018-09-07
Always use estimateGas when computing forgeOrg gas

dkent600:
Resolves: #342 

- always uses `estimateGas` when computing `forgeOrg` gas. No more need to manually compute the gasLimit
- caps gas limit on *any* attempt to override the default gas limit on a transaction
- rewrites the createDaoScript to use `DAO.new`
- uses AbsoluteVote for schemes (other than ContributionReward) in the createGenesisDao script


lib/wrappers/daoCreator.ts
```diff
@@ -99,7 +98,18 @@ export class DaoCreatorWrapper extends ContractWrapperBase {
...
+      options.name,
+      options.tokenName,
+      options.tokenSymbol,
+      options.founders.map((founder: FounderConfig) => web3.toBigNumber(founder.address)),
+      options.founders.map((founder: FounderConfig) => web3.toBigNumber(founder.tokens)),
+      options.founders.map((founder: FounderConfig) => web3.toBigNumber(founder.reputation)),
+      controllerAddress,
+      options.tokenCap];
+
+    // always use estimateGas or will fail when using non-universal controller
```
orenyodfat: "or will fail when using non-universal controller ? why ?

gasLimits.js
```diff
@@ -10,34 +10,16 @@ const arcForgeOrgGasLimit = require("./arcConstants.js").ARC_GAS_LIMIT;
 const gasLimitsConfig =
 {
   /**
-   * The gas limit used by Arc to forge a DAO with foundersInGasLimitArc founders
-   * and to `new` GenesisProtocol.
+   * The gas limit used by Arc to `new` GenesisProtocol.
```
orenyodfat: This comment is not clear.please clarify

gasLimits.js
```diff
@@ -10,34 +10,16 @@ const arcForgeOrgGasLimit = require("./arcConstants.js").ARC_GAS_LIMIT;
...
-  "gasLimit_perFounder": 50000,
-  /**
-   * How many founders are already accounted-for in the gaslimit from Arc
-   */
-  "foundersInGasLimitArc": 3,
   /**
    * Gas limit sufficient (though not necessarily optimal) for all other transactions
    * besides creating DAOs
    */
   "gasLimit_runtime": 4543760,
```
orenyodfat: How this number is calculated ? from where it is taken ?

lib/wrappers/daoCreator.ts
dkent600: It runs out of gas.  I don't know why the default gas limit doesn't work.

gasLimits.js
```diff
@@ -15,7 +15,8 @@ const gasLimitsConfig =
   "gasLimit_arc": arcForgeOrgGasLimit,
   /**
    * Gas limit sufficient (though not necessarily optimal) for all other transactions
-   * besides creating DAOs.  This is taken from the default that truffle puts in truffle.js.
+   * besides creating DAOs and instantiating a new GenesisProtocol.
+   * This is taken from the default that truffle puts in truffle.js.
```
orenyodfat: why did you choose to use truffle default values ? why do we care about truffle default value ?
how did truffle choose this value ?

dkent600: I think the code that uses this may not be required.  It is very old code that actually predates me, and has never, that I recall, been questioned.  I will test with it removed and report back.

BTW I am also looking into removing the other setting, `gasLimit_arc`, a task which is proving to be tricky, but with both of those settings gone we could completely remove some code including the dependency on the Arc test/constants.js file.

dkent600:
@orenyodfat I have removed arcConstants.js and gasLimits.js, so there is no longer a dependency on gas numbers from Arc.

The default gas limit used by truffle is now stored in defaults.json as a global configuration setting.  It is required for certain operations to succeed with enough gas.

I have updated the description in this PR to include all of these recent changes.

config/default.json
```diff
@@ -4,6 +4,7 @@
   "cacheContractWrappers": true,
   "logLevel": 9,
   "estimateGas": false,
+  "defaultGasLimit": 4543760,
```
orenyodfat: why this value and not other (e.g 4543759) ?
Is there a good reason for that except that "truffle use that" ?

lib/contractWrapperBase.ts
```diff
@@ -184,9 +184,7 @@ export abstract class ContractWrapperBase implements IContractWrapper {
 
     const currentNetwork = await Utils.getNetworkName();
 
-    const web3 = await Utils.getWeb3();
-
-    const maxGasLimit = await computeMaxGasLimit(web3);
+    const maxGasLimit = await UtilsInternal.computeMaxGasLimit();
 
     if (currentNetwork === "Ganache") {
       return maxGasLimit; // because who cares with ganache and we can't get good estimates from it
```
orenyodfat: Gas estimation for Ganache should be the same as other networks.
In general, we should strive to execute same arc.js code regardless that networks it is connected to.(as possible as can be done). 
This line should be removed.

lib/contractWrapperBase.ts
```diff
@@ -321,9 +319,9 @@ export abstract class ContractWrapperBase implements IContractWrapper {
       let error;
 
       const gasPriceComputer = ConfigService.get("gasPriceAdjustment") as GasPriceAdjustor;
+      const web3 = await Utils.getWeb3();
 
       if (gasPriceComputer && !web3Params.gasPrice) {
```
orenyodfat: gasPriceComputer -> gasPriceCompute ?

lib/migrations/2_deploy_schemes.ts
```diff
@@ -53,7 +53,7 @@ export const arcJsDeployer = (
     console.log(`Deploying schemes to ${network}`);
 
     const DAOToken = await Utils.requireContract("DAOToken");
-    const gasLimit = await computeMaxGasLimit(web3);
+    const gasLimit = await UtilsInternal.computeMaxGasLimit();
     const gasPrice = 10000000000; // 10 Gwei
```
orenyodfat: const gasPrice = 10000000000; // 10 Gwei ?
from where this come from ?

dkent600: The scripts doubles that in a few specific cases where it doesn't suffice.

lib/contractWrapperBase.ts
dkent600: fixed

lib/contractWrapperBase.ts
dkent600: Ganache returns gas estimates that are not sufficient.  At the time I wrote this code I spent a lot of time on this issue.  I just now gave it another try, but encountered the same result.  I am in the process of moving to truffle 5.0 and web3 1.0 -- maybe something will change then, I can retry when I have things working enough to test.

I think this may be a case requiring network-specific code.

config/default.json
orenyodfat: quote-there-is-nothing-without-reason-gottfried-leibniz-126-13-19-2

orenyodfat: lets find out the reason or set it to max gas limit

lib/contractWrapperBase.ts
orenyodfat: Is this a known ganache issue ? If it is please add a link to this issue.

lib/migrations/2_deploy_schemes.ts
orenyodfat: fix 10 Gwei is a workaround.
please set the appropriate gas price needed.
You can read that from eth gas station ...(alchemy done that)

lib/contractWrapperBase.ts
dkent600: https://github.com/trufflesuite/ganache-core/issues/151

lib/migrations/2_deploy_schemes.ts
dkent600: Yeah, I was thinking of that too.  Would you mind waiting on this until I have Truffle 5 working?  As you know, they have lots of migration changes and I'd like to code this in the context of those changes.  I'm doing the work right now of upgrading to Truffle 5.

lib/contractWrapperBase.ts
orenyodfat: could you please add a link to that on the code ?

lib/migrations/2_deploy_schemes.ts
orenyodfat: please open an issue for that.

dkent600:
New tickets:

https://github.com/daostack/arc.js/issues/358
https://github.com/daostack/arc.js/issues/357
https://github.com/daostack/arc.js/issues/359

@orenyodfat

Issue #351, opened 2018-09-12
be consistent with use of protocol in URIs in the default configuration 


orenyodfat:
https://github.com/daostack/arc.js/blob/master/config/default.json

1. "globalGenTokenAddresses": {
    "default": "0x543Ff227F64Aa17eA132Bf9886cAb5DB55DCAddf",
    "ganache": "0xdcf22b53f327b4f7f3ac42d957834bd962637555"
  },
  Default token address at ganache ? 
  please explain.

2. "providerUrl": "http://127.0.0.1" ..
     why all other network has  "host": "127.0.0.1" ?

{
  "autoApproveTokenTransfers": true,
  "defaultVotingMachine": "AbsoluteVote",
  "cacheContractWrappers": true,
  "logLevel": 9,
  "estimateGas": false,
  "gasPriceAdjustor": null,
  "txDepthRequiredForConfirmation": {
    "default": 7,
    "ganache": 0,
    "live": 20
  },
  "providerPort": 8545,
  "providerUrl": "http://127.0.0.1",
  "globalGenTokenAddresses": {
    "default": "0x543Ff227F64Aa17eA132Bf9886cAb5DB55DCAddf",
    "ganache": "0xdcf22b53f327b4f7f3ac42d957834bd962637555"
  },
  "networkDefaults": {
    "live": {
      "host": "127.0.0.1",
      "port": 8546
    },
    "ganache": {
      "host": "127.0.0.1",
      "port": 8545
    },
    "ropsten": {
      "host": "127.0.0.1",
      "port": 8548
    },
    "kovan": {
      "host": "127.0.0.1",
      "port": 8547
    }
  }
}

dkent600:
@orenyodfat 

> Default token address at ganache ?

The address on ganache is not the same as the address on Kovan and mainnet

> why all other network has "host": "127.0.0.1" ?

Are you thinking it should not?  This is typically where node clients listen.

orenyodfat:
"The address on ganache is not the same as the address on Kovan and mainnet ."
Yea, though how you use same token address on ganache for each run ? by using same deployment account each time ?

dkent600:
> by using same deployment account each time ?

Yes, exactly.

PR #352, opened 2018-09-13
remove "http://" from provider Url

dkent600:
Resolves: #351


dkent600:
@tibetsprague 

The "why" is in the ticket this resolves:  https://github.com/daostack/arc.js/issues/351#issuecomment-420779472

PR #362, opened 2018-10-23
Upgrade to Arc v56

dkent600:
* upgrades to Arc v56
  * uses the new [Infra architecture](https://github.com/daostack/infra)
  * _avatar => _organization in all voting machine events where _avatar appeared
  * Voting machine `propose` functions have a new required argument: `organizationAddress`, which is what will appear as `_organization` in the voting machine events for the new proposal.  This argument is used to link a scheme with an organisation. If this argument is set to zero the _organisation will be the `msg.sender`.
  * `GenesisProtocol.getBoostedProposalsCount` now takes the avatar address in addition to the contract address.
  * Adds `GenesisProtcol.getProposalOrganizationId` that  returns a promise of the `organizationId` of the proposal.
  * It is no longer possible to compute a GenesisProtocol threshold value scoped to an entire DAO if the DAO has more than one scheme using GenesisProtocol to create proposals.
* adds wrappers for all of the reputation-bootstrapping schemes
* fixes bug where manually-submitted gas limit was not being respected
* makes public `contractWrapperFactory.ensureSolidityContract` for obtaining the unhydrated truffle contract, useful with contracts like the reputation-bootstrapping schemes that are wrapped but have not been deployed by Arc.js.
* adds `contractWrapperBase.getDefaultPermissions` for internal use when registering schemes that have a wrapper but no particular permissions other than the ability to access the controller.
* minimizes loading of contract ABI json files, enabling applications to better optimize how they bundle these files which are quite bulky (`ControllerService`).
* caches unhydrated Truffle contracts, so they are only instantiated once.
* adds:
  * `WrapperService.universalSchemeFactories` (`import UniversalSchemeFactories`)
  * `WrapperService.nonUniversalSchemeFactories` (`import NonUniversalSchemeFactories`)
  * `WrapperService.schemeFactories` (`import SchemeFactories`)
* GenesisProtocol parameter `daoBountyConst` validation now uses <= and >= instead of < and > (daoBountyConst must be greater than `stakerFeeRatioForVoters` and less than `2*stakerFeeRatioForVoters`). 

**Breaking**
* avatar => _organization in all voting machine events where _avatar appeared
* Voting machine `propose` functions have a new required argument: `organizationAddress`, which is what will appear as `_organization` in the voting machine events for the new proposal.
* `GenesisProtocol.getBoostedProposalsCount` now takes the avatar address in addition to the contract address.
* It is no longer possible to compute a GenesisProtocol threshold value scoped to an entire DAO if the DAO has more than one scheme using GenesisProtocol to create proposals.
* GenesisProtocol parameter `daoBountyConst` validation now uses <= and >= instead of < and > 
(daoBountyConst must be greater than `stakerFeeRatioForVoters` and less than `2*stakerFeeRatioForVoters`).


orenyodfat:
"Voting machine propose functions have a new required argument: organizationAddress, which is what will appear as _organization in the voting machine events for the new proposal."

Please note that this argument is used to link a scheme with an organisation. If this argument is set to zero the _organisation will be the msg.sender

lib/wrappers/locking4Reputation.ts
```diff
@@ -0,0 +1,474 @@
...
+    super.hydrated();
+    /* tslint:disable:max-line-length */
+    this.Redeem = this.createEventFetcherFactory<Locking4ReputationRedeemEventResult>(this.contract.Redeem);
+    this.Release = this.createEventFetcherFactory<Locking4ReputationReleaseEventResult>(this.contract.Release);
+    this.Lock = this.createEventFetcherFactory<Locking4ReputationLockEventResult>(this.contract.Lock);
+    /* tslint:enable:max-line-length */
+  }
+}
+
+// export class Locking4ReputationType extends ContractWrapperFactory<Locking4ReputationWrapper> {
```
tibetsprague: why leave these comments in?

lib/utilsInternal.ts
```diff
@@ -22,7 +22,25 @@ export class UtilsInternal {
...
    * Returns the last mined block in the chain.
    */
-  public static async lastBlock(): Promise<number> {
+  public static async lastBlock(): Promise<BlockWithoutTransactionData> {
+    const web3 = await Utils.getWeb3();
+    return promisify((callback: any): any => web3.eth.getBlock("latest", callback))() as any;
+  }
+
+  /**
+   * Returns the last mined block in the chain.
```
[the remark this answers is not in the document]
dkent600: fixed

lib/wrappers/locking4Reputation.ts
dkent600: removed

PR #363, opened 2018-10-24
implement arc.js integration with Portis as web3 provider

tkeidar:


lib/utils.ts
```diff
@@ -47,7 +48,7 @@ export class Utils {
    * When called for the first time, web3 is initialized from the Arc.js configuration.
    * Throws an exception when web3 cannot be initialized.
    */
-  public static async getWeb3(): Promise<Web3> {
+  public static async getWeb3(portisConfig?: any): Promise<Web3> {
```
orenyodfat: How about rename 'portisConfig' to 'providerConfig and add 'provideType' param so it could be generalised to any provider ?

dkent600: Thanks for the suggestion, @tkeidar 

This change is not necessary, and we don't want references to particular providers in Arc.js, which should be agnostic to the type of provider. In any case it would not be the right place for it.  The right place would be to add `web3?: Web3` to `InitializeArcOptions` that is passed to `InitializeArcJs`.

In the meantime, what you can do right now is simply inject the desired web3 instance into `global`, prior to calling `InitializeArcOptions`, and Arc.js will pick it up just fine.  This is how the Alchemy caching and arc.js-scripts do it.

For example from arc.js-scripts:

https://github.com/daostack/arc.js-scripts/blob/43c88db5eed72cc37832114191ab4d9d749c82f2/scripts/arcScript.ts#L139

orenyodfat: @dkent600 @tkeidar lets find the best way to enable using Portis as a provider.

PR #368, opened 2018-11-06
add GenesisProtocolWrapper.getThresholdForSchemeAndCreator

dkent600:
* adds `GenesisProtocolWrapper.getThresholdForSchemeAndCreator`
* renames `GenesisProtocolWrapper.getThreshold` to `GenesisProtocolWrapper.getThresholdFromProposal`
* improves some GenesisProtocol documentation

**Breaking**
* renames `GenesisProtocolWrapper.getThreshold` to `GenesisProtocolWrapper.getThresholdFromProposal`

### DIFF
--- a/lib/wrappers/genesisProtocol.ts
+++ b/lib/wrappers/genesisProtocol.ts
@@ -338,7 +338,7 @@ export class GenesisProtocolWrapper extends IntVoteInterfaceWrapper
    * this voting machine by the scheme at `schemeInfo.address` for the given `creatorAddress`
    * (`creatorAddress` is the `organizationAddress` given to the `propose` method),
    * and the GenesisProtocol parameters `thresholdConstA` and `thresholdConstB` which
-   * are registered on behalf of this GenesisProtocol contract with an avatar's Controller,
+   * are registered on behalf of the given scheme with an avatar's Controller,
    * keyed by the given `creatorAddress`.
    *
    * Thus `creatorAddress` must be an avatar address and this function will not work if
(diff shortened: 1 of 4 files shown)


lib/wrappers/genesisProtocol.ts
```diff
@@ -332,13 +333,69 @@ export class GenesisProtocolWrapper extends IntVoteInterfaceWrapper
...
+   * Otherwise you must use `getThresholdFromProposal`.
+   *
+   * Note that there will be a separately-scoped threshold for proposals created for this
+   * voting machine by any other scheme than the one given.
+   *
+   * Note that for Arc.js schemes you can just pass an Arc.js contract wrapper for `schemeInfo`.
+   */
+  public async getThresholdForSchemeAndCreator(
+    schemeInfo: ThresholdSchemeInfo,
+    creatorAddress: Address): Promise<BigNumber> {
```
tibetsprague: is creator the right word here? why not organizationAddress?

lib/wrappers/genesisProtocol.ts
```diff
@@ -332,13 +333,69 @@ export class GenesisProtocolWrapper extends IntVoteInterfaceWrapper
...
+   * The computation depends on the current number of boosted proposals that were created for
+   * this voting machine by the scheme that created the given proposal,
+   * and the GenesisProtocol parameters `thresholdConstA` and `thresholdConstB` which
+   * are stored with the proposal.
+   *
+   * Note that there will be a separately-scoped threshold for proposals created for this
+   * voting machine by any other scheme than the one that created the given proposal.
    */
-  public async getThreshold(proposalId: Hash): Promise<BigNumber> {
+  public async getThresholdFromProposal(proposalId: Hash): Promise<BigNumber> {
```
tibetsprague: "From" seems wrong here too, isn't it "for"?

lib/wrappers/genesisProtocol.ts
dkent600: Because it isn't necessarily an organization.  It can be a contract too.

lib/wrappers/genesisProtocol.ts
dkent600: but I realize that it must in this case be an avatar address (as documented), so I'll change that

dkent600: Done

lib/wrappers/genesisProtocol.ts
dkent600: but I realize that for this particular function to work, it must  be an avatar address (as documented), so I'll change that.

dkent600: done

lib/wrappers/genesisProtocol.ts
tibetsprague: ah interesting and a bit strange but ok

Issue #82, opened 2018-02-12
In ContributionReward.proposeContributionReward, should caller hash the description?


dkent600:
Currently `ContributionReward.proposeContributionReward` hashes the description of the proposal.

Should the caller be responsible for doing the hashing?

If the caller is responsible for managing external storage of the unhashed description, then they may want to have the hash value to use as a key into their storage.  (See `ContributionReward.getDaoProposals` and #116 for how the caller can obtain the hash value.)

But is there any reason why the caller might need to be able to do the hashing themselves?

dkent600:
@leviadam

dkent600:
@tibetsprague What do you think?

tibetsprague:
hmm, do we care how the hashing is done? my intuition is that we should do the hashing in arc.js but im not sure

dkent600:
@leviadam @orenyodfat @tibetsprague  Closing due to lack of objection to the status quo.

bohendo:
Having arc.js handle hashing is blocking progress on my recent [proposal to migrate data to IPFS](https://gitlab.com/bohendo/alchemy/blob/master/proposals/decentralize-proposal-data.md).

In the future, other clients might want to store data in IPFS. Or maybe clients will want to utilize Ethereum's swarm storage once it's ready. The current implementation lacks the flexibility of experimenting with these options.

bohendo:
Don't stress about patching this, since I need to use old (ie currently deployed) contracts, it seems like I need to use a version of arc.js prior to 0.0.0-alpha.85 so I don't think I'd be able to use a new release even if it were available. I suspect I can work around this issue by calling this method manually w/out using the arc.js library.

dkent600:
Reopening to consider @bohendo's issue

PR #382, opened 2018-12-06
Upgrade to Arc v58, Rinkeby contract addresses

dkent600:
Resolves: #82 

* upgrades to Arc v58, including particularly changes to reputation bootstrapping contracts
* adds contract addresses for rinkeby
* Arc.js migration script now maintains the Ganache contract addresses supplied by the DAOstack migration repo, meaning that the account addresses have changed as a result.
* `ContributionReward.proposeContributionReward` will now optionally take a proposal description that has already been hashed (and won't hash it).  See `ProposeContributionRewardParamsdescriptionIsHashed`.
* unstubs stubbed-out GenesisProtocol scheme parameters.
* removes `ContributionReward.ContributionProposal.contributionDescriptionHash`

**Breaking**
* Ganache account addresses have changed.
* removes `ContributionReward.ContributionProposal.contributionDescriptionHash`


lib/wrappers/contributionReward.ts
```diff
@@ -154,7 +155,7 @@ export class ContributionRewardWrapper extends ProposalGeneratorBase {
       eventContext,
       this.contract.proposeContributionReward,
       [options.avatar,
-      Utils.SHA3(options.description),
+      options.descriptionIsHashed ? options.description : Utils.SHA3(options.description),
```
tibetsprague: makes a bit more sense to me to pass in whether you Arc.js to hash or not, so by default it would not be hashed, but this is ok too

lib/wrappers/contributionReward.ts
```diff
@@ -587,7 +587,6 @@ export interface NewContributionProposalEventResult {
 export interface ContributionProposal {
   proposalId: Hash;
   beneficiaryAddress: Address;
-  contributionDescriptionHash: Hash;
```
tibetsprague: why does this get removed?

package.json
```diff
@@ -51,7 +51,7 @@
   },
   "homepage": "https://github.com/daostack/arc.js#readme",
   "dependencies": {
-    "@daostack/migration": "0.0.0-alpha.56-v3",
+    "@daostack/migration": "0.0.0-alpha.58-v2",
```
tibetsprague: backwards?

lib/wrappers/contributionReward.ts
dkent600: @tibetsprague I forgot about that.  I've added it to the breaking changes above.  It was removed from the contract as part of this PR:  https://github.com/daostack/arc/pull/561/files

lib/wrappers/contributionReward.ts
dkent600: @tibetsprague I did it this way to avoid creating a breaking change.  For what it is worth, it has the most convenient option as the default.

package.json
dkent600: @tibetsprague no, the earlier one is 56, the latter, 58

config/default.json
```diff
@@ -18,6 +18,10 @@
       "host": "127.0.0.1",
       "port": 8546
     },
+    "private": {
```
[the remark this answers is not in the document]
dkent600: private is a catchall.  ganache is a special case for a network running with a particular network id that arc.js supplies.  So, for example, the user might have ganache running with a unknown network id, but arc.js can't know that it is in fact ganache, so will name it "private".

BTW, "private" is the name used by the migrations package and web3.

package.json
```diff
@@ -63,7 +66,7 @@
     "ethereumjs-abi": "^0.6.5",
     "ethjs-abi": "0.1.8",
     "fs-extra": "^5.0.0",
-    "ganache-cli": "^6.1.6",
+    "ganache-cli": "^6.2.3",
```
orenyodfat: 6.2.5 ?

dkent600: will do in the Arc upgrade PR

